PR #16, opened 2016-07-02
Fixed typo of generated serializer name

tacoo:
Hi,

Thank you for the great software! I learned a lot about asm!
But one thing I noticed that generated serializer/deserializer uses same name. So if the binary check sum is matched, it might be a problem so I made a small changes in order to use different name.

Thank you


cowtowncoder:
Very good catch -- I agree names should differ, just in case, even if collision is not likely.
Thank you for fixing this.

PR #27, opened 2017-07-27
Fix incorrect class name use in Afterburner.

stevestorey:
When using INVOKEVIRTUAL the first parameter needs to be the internal class
name (i.e. com/fasterxml/x/y/SomeClass) not the type descriptor
(Lcom/fasterxml/x/y/Class;) and some classloaders are more strict about this
than others it seems.

Resolves https://github.com/FasterXML/jackson-module-afterburner/issues/68


cowtowncoder:
Ok sounds good. One question -- is there any known way to trigger this on Oracle JVM, with modified tests? I am fine with the change, but it would be great to guard against regression.

stevestorey:
I will say 'probably' but I'm not sure how - it was only going wrong for me in the presence of the wildfly classloaders, and not with other ones. But I'd have thought (and I'm a novice on asm!) the invokedynamic instruction wouldn't have been valid in the first place on any JVM, unless they were fixing up the Lcom/.. -> com/.. which would be a bit weird?

PR #29, opened 2017-09-03
Replace ASM code-gen with ByteBuddy in afterburner

geoand:
This PR replaces all the parts that contain ASM code-gen logic, with ByteBuddy.

I should note that while replacing the ASM code-gen code with the relevant ByteBuddy code, I tried to encapsulate the common logic and "algorithms" of the various parts into self standing classes that can be composed and reused.

Furthermore, the new code also contains a few minor changes to the generated bytecode in order to remove the following unnecessary instructions:

- When generating code for a single method/field, the index is no longer loaded onto the stack since it's not used
- The IllegalArgumentException is only generated when using the `switch` method. In the case of a single method/field or when the `if` method is used, the exceptions are not generated since it was impossible for control to reach them anyway.
- The bytecode that throws the IllegalArgumentException adds the whole string to the stack and forgoes the use of `StringBuilder`. This is because at the time of the bytecode creation, the whole exception message is already known.

It goes without saying that I am willing to make any changes you deem necessary since I acknowledge that the surface area of the changes is rather large and that I might have missed something and/or not designed/implemented something in the optimal way.

### DIFF
--- a/afterburner/pom.xml
+++ b/afterburner/pom.xml
@@ -15,6 +15,7 @@ field access and method calls
   <url>https://github.com/FasterXML/jackson-modules-base</url>
 
   <properties>
+    <version.butebuddy>1.7.8</version.butebuddy>
     <!-- Generate PackageVersion.java into this directory. -->
     <packageVersion.dir>com/fasterxml/jackson/module/afterburner</packageVersion.dir>
     <packageVersion.package>${project.groupId}.afterburner</packageVersion.package>
@@ -27,7 +28,7 @@ field access and method calls
        Similarly, org.objectweb.asm is shaded... why require?
        (will try to hide via resolution directive)
       -->
-    <osgi.import>org.objectweb.asm;resolution:=optional,
+    <osgi.import>net.bytebuddy;resolution:=optional,
 *
      </osgi.import>
 <!--
@@ -48,9 +49,9 @@ field access and method calls
       <artifactId>jackson-databind</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.ow2.asm</groupId>
-      <artifactId>asm</artifactId>
-      <version>${version.asm}</version>
+        <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy</artifactId>
+        <version>${version.butebuddy}</version>
     </dependency>
     <dependency> <!--  tests use Jackson annoations -->
       <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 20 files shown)


geoand:
For some odd reason it seems like TravisCI is not setting up the JDK 7 tests correctly.

tacoo:
Seems like oraclejdk7 is no longer supported.
So use openjdk7 or drop jdk7 completely? 

https://github.com/travis-ci/travis-ci/issues/7884#issuecomment-308451879

cowtowncoder:
Excellent! Fixing travis builds now.

cowtowncoder:
Ok: I am ready to commit this to `master`, since it's for 3.0.0 now, but looks like there's a conflicting change unfortunately. Would it be easy enough for you to resolve this?

I am hoping to do performance tests after merge: should not show any difference, but just want to verify. I also have some ideas for more code generation that I'd like to explore, giving me a good excuse to see how BB API works.

geoand:
Hi,

Glad to hear it 😀.

Of course I'll update the PR to resolve the conflicts.
I'll probably do that over the weekend.

I'll get back to you once everything is in order

cowtowncoder:
@geoand excellent, thank you in advance!

geoand:
No problem 😀!

geoand:
@cowtowncoder While starting work on the resolving the conflicts, I am failing to build the various modules from source (for example _jackson-core_), because I can't find and build the source for jackson-base.
I am sorry if I am missing something obvious, but I tried searching the entire FasterXML organization and didn't come up with any Maven modules named jackson-base

Thank you in advance!

cowtowncoder:
@geoand Apologies for confusion: `jackson-base` is bit hidden: it comes from `jackson-bom`, which just become multi-maven-project repo. Intent is for `jackson-base` to become parent pom for Jackson component themselves, and strip down `jackson-bom` to be something other projects may import or use as parent, but without getting too much of cruft that only Jackson projects require.

I thought I had deployed 3.0.0-SNAPSHOT of `jackson-base` to Sonatype snapshot repository, but perhaps this has not happened. Or maybe there's chicken and egg problem there.

Anyway, if you locally build `jackson-bom`, that should to the trick.

geoand:
No problem!

Thank you for the pointers! I'll check it out and let you know if anything
comes up.

geoand:
Hi!

I just rebased the PR onto master.
Everything works correctly except one test, namely: `TestUnwrapped.testSimpleSerialize`.
However this test was not working on `master` branch either.

cowtowncoder:
@geoand Thanks: chances are failure may be due to changes in `jackson-databind` that maybe weren't synced (not sure if snapshot builds work right now), I'll verify things work for me, and if not see what is happening.

geoand:
Sounds good! Thanks!

PR #32, opened 2017-09-23
Fix introspector chaining in JaxbAnnotationIntrospector.hasRequiredMarker()

vojtechhabarta:
Let's say I have a property marked with `@JsonProperty(required = true)`. 
I also used `objectMapper.registerModules(ObjectMapper.findModules(classLoader))` to register modules but now the property is not considered "required".

That's because there is a chain of `JaxbAnnotationIntrospector` and `JacksonAnnotationIntrospector` (in this order) and `JaxbAnnotationIntrospector.hasRequiredMarker()` returns `Boolean.FALSE` so `JacksonAnnotationIntrospector` is not called (in `AnnotationIntrospectorPair`).

This PR fixes this case, `JaxbAnnotationIntrospector.hasRequiredMarker()` now returns `null` so property can be processed by subsequent introspector.

This PR also changes that `JaxbAnnotationIntrospector.hasRequiredMarker()` now authoritatively returns `Boolean.FALSE` when annotation is present. This behavior is consistent with `JacksonAnnotationIntrospector`.

### DIFF
--- a/jaxb/src/main/java/com/fasterxml/jackson/module/jaxb/JaxbAnnotationIntrospector.java
+++ b/jaxb/src/main/java/com/fasterxml/jackson/module/jaxb/JaxbAnnotationIntrospector.java
@@ -441,13 +441,10 @@ public class JaxbAnnotationIntrospector
         if ((attr != null) && attr.required()) {
             return Boolean.TRUE;
         }
-        // 09-Sep-2015, tatu: Not 100% sure that we should ever return `false`
-        //   here (as it blocks calls to secondary introspector), but since that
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Looking at code, makes sense. Looks like return values were switched around.
The original concern was probably due to `false` being the default (since annotation values do not have "undefined" or `null` possibility). But returning `Boolean.FALSE` as fallback is wrong no matter what.

One practical note: since I want this to be in 2.9.3, I'll manually merge changes (`master` is for 3.0.0).

PR #52, opened 2018-07-18
Interfaces may have non-abstract methods (since java8)

hrhino:
and those methods may override abstract methods in superinterfaces, so don't skip the `hasConcreteOverride` method just because we're materializing an interface.


cowtowncoder:
@hrhino Thank you for contributing this -- somehow I missed it, but hope look into this now and get it merged in for 2.9.7.

cowtowncoder:
@hrhino Ok, looks good. One last (?) thing: unless I have asked for and received CLA earlier, I'd need that from: 

and the usual way is to print, fill & sign, scan, email to `info` at fasterxml dot com. Only needs to be done once and works for all contributions to all Jackson projects.
If that can be done, I'll merge this fix in 2.9 branch and we are good to go.

Thank you again for the contribution!

cowtowncoder:
Manually merged; changed test slightly as it's not a case of having a getter but rather just regular method of interface.

Issue #54, opened 2018-07-30
Afterburner` SuperSonicBeanDeserializer` does not handle JSON Object valued Object Ids (like json)


headw01:
I am attempting to use something like https://github.com/jsog/jsog-jackson so that a .Net client (through a Jackson-using API endpoint) can process references (with ```"$id"``` as the ```@JsonIdentityInfo``` property value and ```"$ref"``` for the JSON Reference).  When Afterburner is used, this doesn't work because the ```SuperSonicBeanDeserializer.deserializeFromObject``` doesn't check with the ```_objectIdReader``` to see if that should be used.

The databind ```BeanDeserializer``` has a check for the ```_objectIdReader``` (per databind#622 in comment)

    public Object deserializeFromObject(JsonParser p, DeserializationContext ctxt) throws IOException
    {
        /* 09-Dec-2014, tatu: As per [databind#622], we need to allow Object Id references
         *   to come in as JSON Objects as well; but for now assume they will
         *   be simple, single-property references, which means that we can
         *   recognize them without having to buffer anything.
         *   Once again, if we must, we can do more complex handling with buffering,
         *   but let's only do that if and when that becomes necessary.
         */
        if ((_objectIdReader != null) && _objectIdReader.maySerializeAsObject()) {
            if (p.hasTokenId(JsonTokenId.ID_FIELD_NAME)
                    && _objectIdReader.isValidReferencePropertyName(p.getCurrentName(), p)) {
                return deserializeFromObjectId(p, ctxt);
            }

The ```SuperSonicBeanDeserializer``` overrides ```BeanDeserializer.deserializeFromObject``` and doesn't have a similar check.

cowtowncoder:
I believe that issue exists, but what is really needed is matching unit test to verify issue, fix.

cowtowncoder:
Was able to repurpose test from `jackson-databind` to reproduce the issue, will fix as suggested.

PR #58, opened 2018-08-08
Added deserializeWithObjectId to SuperSonicBeanDeserializer.

Janekdererste:
This enables the deserializing cyclic object graphs which were created with @JsonIdentityInfo

### DIFF
--- a/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/deser/SuperSonicBeanDeserializer.java
+++ b/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/deser/SuperSonicBeanDeserializer.java
@@ -1,15 +1,18 @@
 package com.fasterxml.jackson.module.afterburner.deser;
 
-import java.io.IOException;
-import java.util.*;
-
-import com.fasterxml.jackson.core.*;
-
-import com.fasterxml.jackson.databind.*;
-import com.fasterxml.jackson.databind.deser.*;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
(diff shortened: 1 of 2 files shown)


Janekdererste:
This fixes #54

Janekdererste:
Please have a look at the unit tests. I didn't know how to force the usage of SuperSonicBeanDeserializer so I just created a Bean with many properties. Maybe there is a more elegant way to do this?

Janekdererste:
And I think some of your unit tests are broken. They also failed on my machine right after I checked out the project.

Janekdererste:
Sorry for the late response, since I was on vacation. I moved the test for object Ids as recommended.

cowtowncoder:
@Janekdererste Thank you for contributing this! I will see if I can merge that in 2.9 (assuming this is problem with 2.x too, not just 3.0), and will also want to verify I understand the issue.

Janekdererste:
If you have any questions while trying to understand the issue, fell free to ask. I've thought a little bit about this now :-)

cowtowncoder:
@Janekdererste Trying to reproduce the issue with 2.9, but for some reason tests actually pass for me.
So it may be that problem only occurs for `master`, which is for 3.0.0. That would also explain test failures as 3.x codebase has changed a lot and sometimes changes in databind are not fully reflected for extension modules (that is, downstream changes sometimes lag).

However... #54 suggests 2.9.x does have issues. So the first thing is to make sure there is reproduction. I'll see if I can port databind test 622 over and go from there.

cowtowncoder:
Was able to reproduce with a test from databind; manually merged fix as suggested (thank you again for your work). Will be include din 2.9.7 release.

PR #70, opened 2019-01-09
Only coerce scalars if ALLOW_COERCION_OF_SCALARS is enabled

### LINKED ISSUE #69: `ALLOW_COERCION_OF_SCALARS` ignored deserializing scalars with Afterburner
Let's look at deserializing an integer field, for example.

The code path in afterburner ends up here:

https://github.com/FasterXML/jackson-modules-base/blob/24865870f790f25f6b05ca8367b80f590b05db12/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/deser/SettableIntFieldProperty.java#L40

If the value being deserialized is not a VALUE_NUMBER_INT, then the code delegates to `_deserializeInt` which will always attempt to coerce a VALUE_STRING to an integer:

https://github.com/FasterXML/jackson-modules-base/blob/24865870f790f25f6b05ca8367b80f590b05db12/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/deser/OptimizedSettableBeanProperty.java#L242-L264

This code path should only be enabled if [MapperFeature.ALLOW_COERCION_OF_SCALARS](http://fasterxml.github.io/jackson-databind/javadoc/2.9.pr3/com/fasterxml/jackson/databind/MapperFeature.html#ALLOW_COERCION_OF_SCALARS) is enabled.

Similarly, this check should be performed when attempting to deserialize other scalar types like boolean and long.

dansanduleac:
Fixes #69 

This fix is for 2.9, but it should merge without conflicts into 2.10 as well.

### DIFF
--- a/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/deser/OptimizedSettableBeanProperty.java
+++ b/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/deser/OptimizedSettableBeanProperty.java
@@ -186,6 +186,7 @@ abstract class OptimizedSettableBeanProperty<T extends OptimizedSettableBeanProp
         }
 
         if (t == JsonToken.VALUE_NUMBER_INT) {
+            _verifyScalarCoercion(ctxt, p, "boolean");
             // 11-Jan-2012, tatus: May be outside of int...
             if (p.getNumberType() == NumberType.INT) {
                 return (p.getIntValue() != 0);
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Sounds reasonable, but would it be possible to add a matching unit test(s)?
`jackson-databind` has a few tests to copy-paste, possibly:

    src/test/java/com/fasterxml/jackson/databind/struct/ScalarCoercionTest.java

dansanduleac:
@cowtowncoder do you think this is OK to merge now?

cowtowncoder:
@dansanduleac Yes, I'd be happy to merge this. Just one more thing: unless I have asked for it (for any Jackson project), we need CLA. It can be found from:

and usual way is to print it, fill & sign, scan, email. Normally I'd ask email to `info` at fasterxml dot com BUT unfortunately we are having some issues with that so if you can make it shareable  some other way (can one attach document to a gist?), that'd be great.

As soon as I get that I'll go ahead and merge the patch. Once again thank you for submitting this & apologies for the hassle wrt cla.

cowtowncoder:
CLA received, filed, merged the fix. Thank you! This will be in 2.9.9 (and 2.10.0+)

PR #75, opened 2019-02-27
upgrade ASM and maven-shade-plugin to JDK 11 ready versions

ulrichenslin:
Upgraded maven-shade-plugin and ASM to versions that work with both JDK 1.8 and JDK 11
All tests pass on both 1.8 and 11

### DIFF
--- a/afterburner/pom.xml
+++ b/afterburner/pom.xml
@@ -77,6 +77,7 @@ field access and method calls
         <!--  We will shade ASM, to simplify deployment, avoid version conflicts -->
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
+        <version>${version.maven.shade.plugin}</version>
         <executions>
           <execution>
             <phase>package</phase>
--- a/mrbean/pom.xml
+++ b/mrbean/pom.xml
@@ -56,6 +56,7 @@ ${project.groupId}.mrbean.*;version=${project.version}
         <!--  We will shade ASM, to simplify deployment, avoid version conflicts -->
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
+          <version>${version.maven.shade.plugin}</version>
         <executions>
           <execution>
             <phase>package</phase>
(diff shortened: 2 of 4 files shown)


cowtowncoder:
Makes sense, although there should be an easier way: `oss-parent` (grandparent pom of projects) specifies property `version.plugin.shade`, and it should only be necessary to override than in one place here, at main level `pom.xml`. If you could change that, verify that works, I'd be happy to merge.
(also need to roll it forward to 2.10 branch, master; I can take care of that).

cowtowncoder:
@ulrichenslin Actually I _think_ that 2.9, 2.10 and `master` all have these updates (since 2.9.8) for plugins, since `oss-parent` version `34` has them.

However, `asm` version is only upgraded for use by plugins, so I can upgrade version used by component. But to minimize risk, I think I'd want to do that just for 2.10(.0); I hope to get that released selative soon (by end of March 2019).

ulrichenslin:
@cowtowncoder Thanks for reviewing the PR.

I agree that it should be managed from oss-parent and the versions provided in 35 are JDK 11 compatible.  The problem is that 2.9 depends on version 34 of oss-parent,  which has the older version of the shading lib:
```
<version.plugin.shade>2.4.3</version.plugin.shade>
```

If 2.10 will be backwards compatible (as per the release notes : https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.10) with 2.9, then this becomes a non-issue.  We can wait for the 2.10 release.

The reason for concern is ASM pre 7 would fail when reading JDK 11 byte code.  If ASM is only used to write byte code, then it should work, but if it reads byte code compiled with source level 11, the it will fail at runtime.

From our testing, this looks to be the last component that needs to be upgraded to make the whole Jackson echo system JDK 11 compatible.

From your point of view, is there any known items that will not be in 2.10 that will prevent JDK 11 compatibility?

cowtowncoder:
@ulrichenslin Yes, 2.10 should have the usual expectations for backwards compatibility for application code. I think it probably makes sense to consider 2.10.0 as the safest bet on JDK 11, and bit less confusing than having to point to 2.9.9 (and not earlier 2.9 versions).

I am not aware of any remaining issues in 2.10, but I am not using JDK 11 myself yet at work; I do rebuilds with it and probably should enable Travis build (assuming JDK 11 is available).

PR #78, opened 2019-03-19
Fixes #77 - Make the dep on javax.activation optional

### LINKED ISSUE #77: Consider making the dep on javax.activation optional
Jackson is very careful to determine whether javax.activation is needed without it needing to be on the classpath, as evidenced here:

https://github.com/FasterXML/jackson-modules-base/blob/master/jaxb/src/main/java/com/fasterxml/jackson/module/jaxb/JaxbAnnotationIntrospector.java#L666

So could we make javax.activation optional?

mbooth101:
This makes the dep on javax.activation optional. The implication is that, if I don't use DataHandlers then there is no need for javax.activation to be present on the classpath. This is consistent with the comments in [JaxbAnnotationIntrospector#isDataHandler()](https://github.com/FasterXML/jackson-modules-base/blob/master/jaxb/src/main/java/com/fasterxml/jackson/module/jaxb/JaxbAnnotationIntrospector.java#L666)

### DIFF
--- a/jaxb/pom.xml
+++ b/jaxb/pom.xml
@@ -48,6 +48,7 @@ data-binding.
        <groupId>javax.activation</groupId>
        <artifactId>javax.activation-api</artifactId>
        <version>1.2.0</version>
+       <optional>true</optional>
      </dependency>
   </dependencies>
 


cowtowncoder:
That seems reasonable, as long as it is not problematic with JDK 9+ module-info.
I will need to merge this to 2.10, I think.

cowtowncoder:
Manually merged similar change (with "provided").

PR #80, opened 2019-03-20
Fixes #77 - Make the OSGi metadata consider javax.activation as an optional dependency

### LINKED ISSUE #77: Consider making the dep on javax.activation optional

mbooth101:
Takes advantage of the OSGi configuration properties defined in the oss-parent pom.


cowtowncoder:
Ok makes sense (alas, I do not directly use OSGi so it's more difficult for me to see).
Just one more thing: since change went in `2.10` (as part of JDK9+ module work), I'll merge this manually (since it's so small).

PR #85, opened 2019-09-05
Blackbird

stevenschlansker:
This is an import of https://github.com/stevenschlansker/jackson-blackbird targeted at 2.12

This seemed like the best place to put it; `java8` was the next closest but it really is not targeted at Java 8 support in particular.  Let me know if there's a more fitting home.

I'd love to reduce the code duplication from Afterburner, although I do consider it more of a "hard fork" so it's not clear there's a great way to share the logic.  Perhaps we could factor out a library of shared functionality.

See also: https://github.com/FasterXML/jackson-bom/pull/26


cowtowncoder:
Just thought I'd drop a note saying that I would actually be interested in seeing something similar as option within `jackson-databind` itself, if feasible. Something to opt-in that would replace accessors for serialization, deserialization. Not sure if this would be practical in 2.12 timeline or not.
Also obviously interested in knowing how performance is, given blackbird exists as a thing and can be benchmarked.

cowtowncoder:
@stevenschlansker Actually, even having blackbird as is as module (for "Afterburner for post-JDK8") would be great. But would probably want to do that for 2.12 branch -- how complicated would it be to rebase?

stevenschlansker:
Very interesting!  I will look into rebasing this weekend or soon after.  It should not be too difficult although I have not looked closely at the changes for 2.12

cowtowncoder:
@stevenschlansker I don't think there are too many Afterburner changes at least, so hopefully it should  be easy.

stevenschlansker:
Hi Tatu, just following up on this - I haven't forgotten, just been busy.
I like your idea of trying to just put it in -databind; I actually think
that makes the code a *lot* simpler.
But I'll need to rework / play with it.

I see 2.12 is released.  Should I still work off 2.12 branch or just target
2.13 at this point?

cowtowncoder:
I know the feeling. :)

On 2.12: that has not yet been released, and the first release candidate will still take until early September I think; likely official 2.12.0 no earlier than early October. I also coordinate changes wrt big features so it all depends on your time (and maybe mine).

I am ok with waiting for 2.13, given that 2.12 has a tons of new features and maybe stability-wise it'd be good. But I am open to either approach.

stevenschlansker:
ok @cowtowncoder this branch is rebased on 2.12!  I will now re-review the changes, try to spruce up my benchmarks, and maybe make some readme updates -- but please let me know if anything else should be done before merge :)

cowtowncoder:
Excellent. I need to complete my work wrt JDK 14 Record support, implied fields polymorphic deser (both contributions) as this is very interesting.

stevenschlansker:
Got some encouraging performance data, especially when using Graal. Wrote it up in the updated readme.
It's a slight-to-moderate performance boost over Vanilla Jackson, and a comparable drop-in to existing Afterburner.

cowtowncoder:
Very cool.

cowtowncoder:
Finally got through backlog, merged this to be in 2.12.0-rc2 for some more user testing!

PR #101, opened 2020-08-26
Adding ppc64le architecture support

kishorkunal-raj:
Hi,
I had added ppc64le support and looks like its been successfully added. I believe it is ready for the final review and merge.

The travis ci build logs can be verified from the link below.
https://www.travis-ci.com/github/kishorkunal-raj/jackson-modules-base/builds/181457864
Please have a look.

Thanks!!"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ jdk:
   - openjdk8
   - openjdk11
 
+arch:
+  - ppc64le
+  - amd64
+
 # allow-list to prevent build of all branches
 branches:
   only:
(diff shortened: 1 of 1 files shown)


cowtowncoder:
@kishorkunal-raj Could you remind of the specific need for this?

Also curious as to details: why is specific version of Maven needed as override here?

Siddhesh-Ghadi:
> Also curious as to details: why is specific version of Maven needed as override here?

There is no specific version needed, its just that there is no default maven package present for ppc64le on travis and we need to manually install it. That being said, we are working on getting maven added has default package on travis for ppc64le so that there is not need to install it manually.

gerrith3:
This package is included in the ppc64le versions of RHEL and Ubuntu - this allows the top of tree to be tested continuously as it is for Intel, making it easier to catch any possible regressions on ppc64le before the distros begin their clones and builds.  This reduces the work in integrating this package into future versions of RHEL/Ubuntu.

cowtowncoder:
Ok thank you -- presumably problems with Afterburner, primarily?
I am mostly curious about actual observed issues, given that this a Java project.

cowtowncoder:
@Siddhesh-Ghadi ah. That makes sense (wrt inclusion lacking). I can add this for now, but if situation changes would be nice to remove installation instruction (although maybe check to see version exists means it does not add much to build time).

PR #107, opened 2020-09-20
Fix detection of inherited default method in Java 8+ interface

cowtowncoder:
## Problem

When the type to materialize extends/implements some interface that has a default method implementation (with a non-bean method name), MrBean will treat the method as if it lacks an implementation.

## Solution

Use Class.getMethod to locate the exact method that the implemented type will utilize, which properly handles this case.


cowtowncoder:
Ok, no, can't backport quite this way (was meant against master)

PR #149, opened 2021-09-02
Fix flaky

alany9552:
The test **TesTxMLLd2.testIdWithJaxbRules** compares the result created by **writeValueAsString()** with a hardcoded string. It turns out the the string returned by  **writeValueAsString()**  is not a deterministic string. The reason is that the type of argument taken by **writeValueAsString()** ObjectMapper. JSON serialization is not required to be deterministic. The solution I've chosen to fix this is to get rid of **JsonMapper.builder()** when creating a new instance, I change the funtion to **newObjectMapper()**. According to Jackson Annotation Documentation, in this case, we can add keywords **@JsonPropertyOrder({ "id", "username","email","department" })** before defining the class **User**. Thus the properties order will be fixed. Following test is failed during the flaky test:

-  **TesTxMLLd2.testIdWithJaxbRules**

This PR helps to guarantee the order of the string returned by writeValueAsString() and make it dererministic.

### DIFF
--- a/jaxb/src/test/java/com/fasterxml/jackson/module/jaxb/id/TestXmlID2.java
+++ b/jaxb/src/test/java/com/fasterxml/jackson/module/jaxb/id/TestXmlID2.java
@@ -137,11 +137,12 @@ public class TestXmlID2 extends BaseJaxbTest
     }
     
     public void testIdWithJaxbRules() throws Exception
-    {
-        ObjectMapper mapper = ObjectMapper.builder()
-        // but then also variant where ID is ALWAYS used for XmlID / XmlIDREF
-                .annotationIntrospector(new JaxbAnnotationIntrospector())
-                .build();
+    {   
+        ObjectMapper mapper = new ObjectMapper();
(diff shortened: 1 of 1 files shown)


jaxb/src/test/java/com/fasterxml/jackson/module/jaxb/id/TestXmlID2.java
```diff
@@ -45,9 +46,10 @@ public void setEmployees(List<User> employees) {
     
     @XmlRootElement(name = "user")
     @XmlAccessorType(XmlAccessType.FIELD)
+    @JsonPropertyOrder({ "id", "username","email","department" })
     static class User
     {
-        @XmlElement @XmlID
+        //@XmlElement @XmlID
```
cowtowncoder, line 52: No, `@XmlID` needs to remain doesn't it?

cowtowncoder:
Thank you @alany9552! I had to tweak this a bit to target 2.13 instead of master, but did as per your suggestion (add annotation to ensure ordering is stable).

PR #150, opened 2021-09-04
fix flaky test testNoneAccessWithXmlElements

RayZhang001:
Fixed flaky test `testNoneAccessWithXmlElements`

The test `testNoneAccessWithXmlElements` compares the result of `MAPPER.writeValueAsString(input)` with a hardcoded string. However, the function `com.fasterxml.jackson.databind.ObjectMapper.writeValueAsString()` would not always return JSON properties in the same order.


cowtowncoder:
Makes sense, thank you @RayZhang001

Issue #159, opened 2021-12-07
NoCtorDeserModule missing - Do you mean NoCtorModule ?


hreinn91:
jackson-modules-base/no-ctor-deser/README.md dictates

`ObjectMapper mapper = JsonMapper.builder() // or mapper for other dataformats
    .addModule(new NoCtorDeserModule())
    // add other modules, configure, etc
    .build();`

But this module is missing from the package as far as I can see. Do you perhaps mean **NoCtorModule** or am I missing something?

hreinn91:
Possible error in 
` jackson-modules-base/no-ctor-deser/src/moditect/module-info.java `

Which references a NoCtorDeserModule which does not infact exisits. Only NoCtorModule exists.

cowtowncoder:
Ugh. This should teach me not to hurry with inclusion and take time to double-check contributions :-(

Thank you for reporting this.

Usually I'd fix this in next minor version (2.14), but in this case I think I'll try to do combination of deprecation for incorrectly named module & addition of proper one in a patch (2.13.1).

PR #160, opened 2021-12-07
159: rename module name to match export

hreinn91:
Renamed NoCtorModule to NoCtorDeserModule


hreinn91:
This is due to issue: https://github.com/FasterXML/jackson-modules-base/issues/159

cowtowncoder:
Thank you for the PR! I will do this slightly differently to go in 2.13.1 (normally I'd wait until patch but in this case seems better to add deprecation markers etc); idea is the same but I'll close this PR.

Thank you once again pointing out this embarrassing flaw :)

Issue #138, opened 2021-06-16
Blackbird doesn't work on Java 15+


jirkapinkas:
On OpenJDK 15 & 16 Blackbird throws this exception. I don't use Module path.

`java.lang.invoke.LambdaConversionException: Exception defining lambda proxy class
	at java.base/java.lang.invoke.InnerClassLambdaMetafactory.generateInnerClass(InnerClassLambdaMetafactory.java:413) ~[na:na]
	at java.base/java.lang.invoke.InnerClassLambdaMetafactory.spinInnerClass(InnerClassLambdaMetafactory.java:304) ~[na:na]
	at java.base/java.lang.invoke.InnerClassLambdaMetafactory.buildCallSite(InnerClassLambdaMetafactory.java:218) ~[na:na]
	at java.base/java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:328) ~[na:na]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.createProperty(BBSerializerModifier.java:151) ~[jackson-module-blackbird-2.12.3.jar:2.12.3]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.lambda$findProperties$0(BBSerializerModifier.java:64) ~[jackson-module-blackbird-2.12.3.jar:2.12.3]
	at com.fasterxml.jackson.module.blackbird.util.Unchecked.lambda$runnable$0(Unchecked.java:31) ~[jackson-module-blackbird-2.12.3.jar:2.12.3]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.findProperties(BBSerializerModifier.java:65) ~[jackson-module-blackbird-2.12.3.jar:2.12.3]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.changeProperties(BBSerializerModifier.java:49) ~[jackson-module-blackbird-2.12.3.jar:2.12.3]
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.constructBeanOrAddOnSerializer(BeanSerializerFactory.java:407) ~[jackson-databind-2.12.3.jar:2.12.3]
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.findBeanOrAddOnSerializer(BeanSerializerFactory.java:290) ~[jackson-databind-2.12.3.jar:2.12.3]
[...]

PR #164, opened 2022-01-12
Blackbird: backport #138 to 2.13

stevenschlansker:


cowtowncoder:
Was about to ask whether you think this is safe... and then realized that your track record is so good that I really need to assume you have considered this and that it is. So will merge for 2.13.2 :)

stevenschlansker:
I'm flattered ;) But I will make a mistake eventually, so please continue to be appropriately skeptical of my contributions! I did test this one pretty thoroughly and a user reported it works well for them, so I feel pretty good about it, despite how weird it looks.

I am following up on OpenJDK core-libs-dev ML to try to understand if we need this workaround forever.

Issue #175, opened 2022-05-24
`jackson-module-jakarta-xmlbind-annotations` should use a Jakarta namespaced Activation API


jamezp:
Currently the `com.fasterxml.jackson.module:jackson-module-jakarta-xmlbind-annotations` uses uses `jakarta.activation:jakarta.activation-api:1.2.2`. This is a `javax` namespaced API. The version 2.0.1 should be used to align with `jakarta.xml.bind:jakarta.xml.bind-api:3.0.1`.

If this can also be backported to 2.13 please let me know as this is a blocker for RESTEasy and WildFly to be able to upgrade.

cowtowncoder:
Ugh. Apologies for the mess here. It does sound like something we'd want for 2.13 too. I'll add a note on PR.

jamezp:
Excellent. Thanks for the quick turnaround on this! Should this be closed now since #177 is merged?

cowtowncoder:
Yes, forgot to close.

cowtowncoder:
Probably not quite around the corner: hoping to finalize 2.14 soon but it'll take a while.
And 2.13.3 just got out so there's not much need for immediate 2.13.4.

PR #176, opened 2022-05-24
[issue172] Migrate to a Jakarta namespaced activation API.

jamezp:
Resolves #175 

I'm happy to back port this to 2.13 if that is acceptable as well.

Issue #152, opened 2021-10-01
jakarta.activation version conflict in 2.13


mkpaz:
I've tried to upgrade to 2.13 branch just recently to get rid of old jaxb namespaces (https://github.com/FasterXML/jackson-modules-base/pull/116). However after I've tried to build my app with `jlink` I've got this:

```
Error: Two versions of module jakarta.activation found in ..\target\modules (jakarta.activation-api-1.2.1.jar and jakarta.activation-2.0.1.jar)
``` 

I'm using new Jakarta XML Bind module, here is dependency tree fragment:

```
[INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.13.0:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.0:compile
[INFO] |  |  \- com.fasterxml.jackson.core:jackson-core:jar:2.13.0:compile
[INFO] |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-properties:jar:2.13.0:compile
[INFO] |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-xml:jar:2.13.0:compile
[INFO] |  |  +- org.codehaus.woodstox:stax2-api:jar:4.2.1:compile
[INFO] |  |  \- com.fasterxml.woodstox:woodstox-core:jar:6.2.6:compile
[INFO] |  +- com.fasterxml.jackson.module:jackson-module-jakarta-xmlbind-annotations:jar:2.13.0:compile
[INFO] |  |  +- jakarta.xml.bind:jakarta.xml.bind-api:jar:3.0.1:compile
[INFO] |  |  |  \- com.sun.activation:jakarta.activation:jar:2.0.1:compile
[INFO] |  |  \- jakarta.activation:jakarta.activation-api:jar:1.2.1:compile
[INFO] |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.13.0:compile
[INFO] |  |  \- org.yaml:snakeyaml:jar:1.28:compile
```

So, `jakarta.activation-api:jar:1.2.1` (Jackson dependency) conflicts with `jakarta.activation:jar:2.0.1` (Jakarta XML Bind transitive dependency), because they're both using same module name. 

To fix this you probably should either exclude `jakarta.activation-api` from module dependencies or switch to the 4.x branch that doesn't depend on old activation API. Links to compare:

[...]

PR #177, opened 2022-05-24
[issue172] Migrate to a Jakarta namespaced activation API.

jamezp:
Upstream is #176, however this one is a bit different. #152 seemed to partially solve this, but never changed the package names. I noticed it imports the implementation too instead of the API, which might be required but I'm not sure. The 2.13 won't currently compile without this change.

Let me know if the API vs the implementation is preferred and I can either change this PR or the upstream #176 one.

### DIFF
--- a/jakarta-xmlbind/pom.xml
+++ b/jakarta-xmlbind/pom.xml
@@ -55,9 +55,9 @@
 
     <!-- 15-Jan-2022, tatu: as per [modules-base#152] -->
     <dependency>
-      <groupId>com.sun.activation</groupId>
-      <artifactId>jakarta.activation</artifactId>
-      <version>2.0.1</version>
+      <groupId>jakarta.activation</groupId>
+      <artifactId>jakarta.activation-api</artifactId>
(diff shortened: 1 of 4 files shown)


cowtowncoder:
I think API should be preferred, not implementation, assuming that's where JAXB/Jakarta-XML-bind annotations live: module makes no use of actual implementation. Unless unit tests used it? (if so, impl should be test-only dependency).

However... I wonder if there are issues wrt backwards compatibility within minor version, wrt impl vs api.
But the main reason I suspect it's fine is that "Jakarta" variants were only added in 2.13 (as per https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.13) so there isn't much usage; and it is likely that anyone trying to use the module might have hit this same issue.

So I will rely on your expertise @jamezp here: I assume your assessment is correct.

jamezp:
Okay perfect. I think the API is appropriate here as well. I've updated this PR to use the API and I'll close the 2.14 one. Thank you!

jamezp:
resolves #175

Issue #174, opened 2022-05-24
Question for Afterburner OptimizedBeanPropertyWriter


BOFA1ex:
[QUESTION]
> why using `broken` and `fallbackWriter` to invoke serializeAsField.
```java
value = this._propertyAccessor.xxxGetter(bean, this._propertyIndex);
JsonSerializer<?> ser = this._serializer;
ser.serialize(value, gen, prov);
```

Before Afterburner module loading with beanSerializerModifier.
```java
public class AfterburnerModule extends Module implements Serializable {
    public void setupModule(Module.SetupContext context) {
        context.addBeanSerializerModifier(new SerializerModifier(cl));
    }
}
```
There was another beanSerializerModifier declare#changeProperties, like below.
```java
 @Override
  public List<BeanPropertyWriter> changeProperties(SerializationConfig config, BeanDescription beanDesc, List<BeanPropertyWriter> beanProperties) {
    for (BeanPropertyWriter property : beanProperties) {
        if (registerJavaTypes.contains(property.getType().getRawClass())) {
            continue;
        }
        property.assignSerializer(shapeSerializerMap.get(ofNullable(property
                .findPropertyFormat(config, beanDesc.getBeanClass()))
                .map(JsonFormat.Value::getShape)
                .orElse(ANY))
        );
    }
    return beanProperties;
}
```
it will be `set broken = true` on `OptimizedBeanPropertyWriter`, like below
```java
public void assignSerializer(JsonSerializer<Object> ser) {
    super.assignSerializer(ser);
    if (this.fallbackWriter != null) {
        this.fallbackWriter.assignSerializer(ser);
    }

    if (!SerializerUtil.isDefaultSerializer(ser)) {
        this.broken = true;
    }
}
```
and then jump to invoke `this.fallbackWriter.serializeAsField(bean, gen, prov);`
using `reflect#invoke` or not `accessor#invoke`.
```java
if (this.broken) {
  this.fallbackWriter.serializeAsField(bean, gen, prov);
} else {
  Object value;
  try {
      value = this._propertyAccessor.objectGetter(bean, this._propertyIndex);
  } catch (Throwable var8) {
      this._handleProblem(bean, gen, prov, var8, false);
      return;
  }

  if (value == null) {
      if (this._nullSerializer != null) {
          gen.writeFieldName(this._fastName);
          this._nullSerializer.serialize((Object)null, gen, prov);
      } else if (!this._suppressNulls) {
[...]

PR #178, opened 2022-05-25
Afterburner: #174 handle primitive/string property with ser

BOFA1ex:
link #174 
Using `serializer` to handle it like `ObjectFieldPropertyWriter`/`ObjectMethodPropertyWriter`

### DIFF
--- a/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/ser/BooleanFieldPropertyWriter.java
+++ b/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/ser/BooleanFieldPropertyWriter.java
@@ -59,7 +59,11 @@ public final class BooleanFieldPropertyWriter
         }
         if (!_suppressableSet || _suppressableBoolean != value) {
             gen.writeFieldName(_fastName);
-            gen.writeBoolean(value);
+            if (_serializer != null) {
+                _serializer.serialize(value, gen, prov);
+            } else {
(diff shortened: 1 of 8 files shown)


cowtowncoder:
Um, no. What problem is this supposed to fix? It adds overhead and seems against the idea of avoiding unnecessary checks. Optimized `BeanPropertyWriter`s should only be assigned when there is only the Jackson standard implementation and we can fully optimize handling.

Under what conditions would this change be needed?

I'd at least need to see a test of something that does not work without change presented.

cowtowncoder:
Um, no. What problem is this intended to fix? It adds overhead and seems against the idea of avoiding unnecessary checks. Optimized `BeanPropertyWriter`s should only be assigned when there is only the Jackson standard implementation and we can fully optimize handling.

Under what conditions would this change be needed?

I'd at least need to see a test of something that does not work without change presented.

BOFA1ex:
Before `afterburner` using `SerializerModifier` to change `BeanPropertyWriter`, there have another modifier to assign serializer in it. and then the serializer is declare with `@JacksonStdImpl`, it will skip the check (`SerializerUtil.isDefaultSerializer(bpw.getSerializer()`). 

But primitive/string method writer do nothing after `access#getValue` just `gen.writeString(value)`.
if not declare with `@JacksonStdImpl`, it using reflect not asm, so why did i try to append `AfterBurnerModule` to improve pref.

Issue #186, opened 2022-08-14
afterburner tests fail for JDK 17


pjfanning:
The CI build builds for Java 8 (ok), Java 11 (small group of failures), Java 17 (lots of afterburner fails -- possibly more issues, the build stops after the afterburner fails).

```
[ERROR] Failures: 
[ERROR]   MyClassLoaderTest.testLoadAndResolveTryParentSameClassTwice:33 first class should be loaded with parent class loader expected:<jdk.internal.loader.ClassLoaders$AppClassLoader@531d72ca> but was:<com.fasterxml.jackson.module.afterburner.util.MyClassLoader@46d59067>
[ERROR]   MyClassLoaderTest.testLoadAndResolveTryParentSameClassTwiceTwoThreads:71 first class should be loaded with parent class loader expected:<jdk.internal.loader.ClassLoaders$AppClassLoader@531d72ca> but was:<com.fasterxml.jackson.module.afterburner.util.MyClassLoader@5276d6ee>
[ERROR] Errors: 
[ERROR]   AfterburnerModuleJDKSerializabilityTest.testMapperWithoutUse:23->_serDeserPointWith:56 » IllegalAccess
[ERROR]   TestAccessFallback.testDeserializeAccess:65 » IllegalAccess failed to access c...
[ERROR]   TestUnwrapped.testUnwrappedDeserialize:51 » IllegalAccess failed to access cla...
[ERROR]   BasicDeserializeTest.testBigNonVoidProperty:300 » IllegalAccess failed to acce...
[ERROR]   BasicDeserializeTest.testBooleans:318 » IllegalAccess failed to access class c...
[ERROR]   BasicDeserializeTest.testEnumField:228 » IllegalAccess failed to access class ...
[ERROR]   BasicDeserializeTest.testFluentMethod:352 » IllegalAccess class com.fasterxml....
[ERROR]   BasicDeserializeTest.testIntMethod:174 » IllegalAccess failed to access class ...
[ERROR]   BasicDeserializeTest.testLongField:214 » IllegalAccess failed to access class ...
[ERROR]   BasicDeserializeTest.testMixed:259 » IllegalAccess failed to access class com....
[ERROR]   BasicDeserializeTest.testMultiIntMethod:179 » IllegalAccess failed to access c...
[ERROR]   BasicDeserializeTest.testNonVoidProperty:285 » IllegalAccess failed to access ...
[...]

PR #198, opened 2023-02-03
fix failing tests in java17 CI run

pjfanning:
Relates to https://github.com/FasterXML/jackson-modules-base/issues/186

See https://github.com/FasterXML/jackson-modules-base/actions/runs/4039573403

Unfortunately, this change stops the gradle-module-plugin working. It does appear to fix the tests. I tried reworking how that gradle-module-plugin is declared but nothing seems to work.

https://github.com/FasterXML/jackson-modules-base/actions/runs/4080247967/jobs/7032511522


cowtowncoder (approved):

cowtowncoder:
Sounds good, thank you for this.

On Gradle module plugin... is that the gradle module metadata thing? It should only be needed with JDK 8 build (for releases).

pjfanning:
> Sounds good, thank you for this.
> 
> On Gradle module plugin... is that the gradle module metadata thing? It should only be needed with JDK 8 build (for releases).

Yes - that plugin.

I simply can't get that plugin to work - I either end up with it not working (the .module files it is supposed to produce do not get produced) or the build just fails.

It may require us to choose - support that funky plugin or be able to run the tests in JDK 17.

In the end, the .module file is only a minor nicety - Gradle is well able to use poms, why we need to have the pom and have this weird .module too - I struggle to see the real benefit.

cowtowncoder:
> > Sounds good, thank you for this.
> > On Gradle module plugin... is that the gradle module metadata thing? It should only be needed with JDK 8 build (for releases).
> 
> Yes - that plugin.
> 
> I simply can't get that plugin to work - I either end up with it not working (the .module files it is supposed to produce do not get produced) or that plugin ends up causing the build to fail.
> 
> It may require us to choose - support that funky plugin or be able to run the tests in JDK 17.
> 
> In the end, the .module file is only a minor nicety - Gradle is well able to use poms, why we need to hack the pom and have this weird .module too - I struggle to see the real benefit.

Yeah I hear you. At the same time, dropping something that was working up until now (for couple of versions) does not sound good either. I do not know how valuable module info files are; I got the impression they might be more of an experimental feature.

But let's see if we can get help: @jjohannes I think you know about this module, right?

jjohannes:
Thanks for letting me know. I am certainly interested in fixing this in the plugin, if it wrongly throws this error in certain cases.

As a first step, can you upgrade the plugin to `0.3.0` (latest version). Maybe that helps already. Can we do that just on this project first to test? It's probably not that easy to update the parent POM where the version is in (?)

Not sure how the different Java version can be responsible for this, but I'll investigate,

pjfanning:
@jjohannes thanks for the input - v0.3.0 still has same problem

jjohannes:
So I don't know why this only fails in the Java 17 build on CI, but this is a valid problem with the build.

_jackson-module-afterburner_ uses the maven shade plugin which generates a "new POM" that is then published. This POM no longer contains the comment. You can see the path in the error message:

`/home/runner/work/jackson-modules-base/jackson-modules-base/afterburner/dependency-reduced-pom.xml`

This is already broken. If you look at he published POM, you see that it also misses the comment:
https://repo1.maven.org/maven2/com/fasterxml/jackson/module/jackson-module-afterburner/2.14.2/jackson-module-afterburner-2.14.2.pom

The real fix would be to add the comment to the `dependency-reduced-pom.xml` as well. Do you know if and how that can be done in the _maven shade plugin_?

pjfanning:
> So I don't know why this only fails in the Java 17 build on CI, but this is a valid problem with the build.
> 
> _jackson-module-afterburner_ uses the maven shade plugin which generates a "new POM" that is then published. This POM no longer contains the comment. You can see the path in the error message:
> 
> `/home/runner/work/jackson-modules-base/jackson-modules-base/afterburner/dependency-reduced-pom.xml`
> 
> This is already broken. If you look at he published POM, you see that it also misses the comment: https://repo1.maven.org/maven2/com/fasterxml/jackson/module/jackson-module-afterburner/2.14.2/jackson-module-afterburner-2.14.2.pom
> 
> The real fix would be to add the comment to the `dependency-reduced-pom.xml` as well. Do you know if and how that can be done in the _maven shade plugin_?

in fairness, XML comments are regarded as nice but non-essential - tools keep them or lose them but hardly anyone makes that behaviour controllable

jjohannes:
> in fairness, XML comments are regarded as nice but non-essential - tools keep them or lose them but hardly anyone makes that behaviour controllable

Sure. One can debate if it was the best decision to use a comment like this by Gradle.

But no matter if one likes it or not, that's how it works.

So if we want to fix it, we need to add that comment to the pom.

In any case I think this is **completely unrelated to this PR**. The build is already broken on `main`. And the problem is probably there for longer and the Maven CI build just did not catch it.

jjohannes:
@pjfanning can you remove the plugin version upgrade commit again from this PR?

@cowtowncoder can you integrate it as it is, as the build failure is not new from this PR?

Then we can open a separate issue for this. If I find some time next week, I can check how this could be fixed.

cowtowncoder:
Ok sounds good, I can merge this.

pom.xml
```diff
@@ -78,36 +78,29 @@ not datatype, data format, or JAX-RS provider modules.
...
           <configuration>
             <excludes>
               <exclude>com/fasterxml/jackson/**/failing/*.java</exclude>
             </excludes>
           </configuration>
-	</plugin>
+        </plugin>
       </plugins>
     </pluginManagement>
-
```
cowtowncoder, line 104: Hmmh. Wait, what, this will remove module metadata from all components, not just Afterburner/Blackbird.
I don't think we want that as it should work fine for others (JAXB annotations, mr-bean, paranamer etC).

jjohannes: Yes. Sorry that seems to be a misunderstanding. What I meant to say is, don't touch this part at all ins this PR as it is unrelated. (don't remove, don't change the version)

jjohannes: I created #200 - you can assign that to me and I will follow up.

Here, you should roll back the change to the root 'pom.xml' completely before merging.

pjfanning: * the only way to get the CI build to pass right now is to remove the gradle module plugin - the jdk17 build fails 100% of the time because of it
* we can add the gradle module plugin back when we work out how to get it work with the dependency-reduced-pom.xml

jjohannes: By chance I just talked to someone with more Maven knowledge. We found this solution (I'll post more details later on the other issue). If you want to have this fixed properly in this PR already you can do this:

https://github.com/FasterXML/jackson-modules-base/issues/200#issuecomment-1416711529

(instead of removing the gradle metadata plugin completely in this repo)

pjfanning: reverted this bit and applied @jjohannes fix

cowtowncoder (approved):

cowtowncoder:
Okay looks good; will merge, need to manually push to 2.15 and master as well.

I wonder if we can have downstream tests to verify successful publishing of GMM (ditto for OSGi metadata). But first things first.

cowtowncoder:
Merged to 2.15 and master as well.

jjohannes:
> I wonder if we can have downstream tests to verify successful publishing of GMM (ditto for OSGi metadata). But first things first.

The test I added here was supposed to inform about issues like this:
https://github.com/FasterXML/jackson-integration-tests/blob/master/src/test/java/com/fasterxml/jackson/integtest/gradle/GradleTest.java

But it's not "strict" enough. I will have a look. I should also upgrade the Gradle version there to make `jackson-integration-tests` work on Java 17.

jjohannes:
https://github.com/FasterXML/jackson-integration-tests/pull/16

Issue #126, opened 2021-01-20
jackson-module-jaxb-annotations 2.11.2 has wrong module info (Java 9 instead of Java 11)


oleg-nenashev:
Looks like there are some minor issues in jackson-module-jaxb-annotations 2.11.2 multi-release JAR.
It looks like Java 9 - focused class is used instead. Not a big deal, but there is a warning generated by Maven Enforcer Plugin.

```
[INFO] --- maven-enforcer-plugin:3.0.0-M3:enforce (display-info) @ MY_PROJECT ---
[INFO] Ignoring requireUpperBoundDeps in com.google.guava:guava
[INFO] Adding ignore: module-info
[WARNING] Invalid bytecodeVersion for com.fasterxml.jackson.module:jackson-module-jaxb-annotations:jar:2.12.1:compile : META-INF/versions/11/module-info.class: expected 55, but was 53
```

cowtowncoder:
I think this is due to improvements by @GedMarc -- Java 9 is not a LTS so we probably don't care about it,  but seems odd there'd be complaint by Maven enforcer plug-in. Maybe there's need to upgrade version of that plugin?

GedMarc:
# closable

timja:
This also happens on java 11 for the record

> Invalid bytecodeVersion for com.fasterxml.jackson.module:jackson-module-jaxb-annotations:jar:2.12.2:compile : META-INF/versions/11/module-info.class: expected 55, but was 53

cowtowncoder:
If anyone has ideas of how to address that, I'm all ears. But as of now I don't know what, if anything, to do.
Seems like some verifier is expecting bytecode to be compiled targeting Java 11, but it was Java 9. Not sure why that would be considered harmful (reverse would be).

GedMarc:
arg, we can move it to version/9 instead of 11

it is a setting in moditect

GedMarc:
but 9 and 10, really shouldn't be used anywhere

GedMarc:
looks like an enforcer minimum jdk version setting actually - 

just update enforcer to allow versions 9 and up, otherwise we'll compile module-info to version 11

GedMarc:
i still feel java 9 is the right version of the module-info.class to allow for everyone, 
@timja Why do have enforcer set for a minimum JDK version for items in META-INF?

GedMarc:
For JDK 11 your minimum allowable java version should be 8, version 7 is no longer compilable -
[...]

PR #199, opened 2023-02-03
jaxb and jakarta-xmlbind put module-info in versions/11

pjfanning:
* this appears wrong and I checked the 2.15.0-SNAPSHOT jars to validate that the module-info.class ends up in versions/11 instead of versions/9
* https://github.com/FasterXML/jackson-modules-base/issues/126


cowtowncoder:
Ok yes; has to be either incomplete change, or problem with merging (`pom.xml` changes do not merge cleanly across branches).

I think I'll actually merge the same thing for `2.14` after merging.

PR #215, opened 2023-07-14
Blackbird: add memory 'leak' OOM warning

### LINKED ISSUE #147: Metaspace/class leak using blackbird
In a standard application, we see class loading level out around ~10-20k classes, however after applying the blackbird module we've seen our loaded classes metric grow until it reaches ~2 million and the JVM is restarted due to an OME. This is running on java 15, although I don't know if it's exclusive to java 15, or which codepath is creating the classes, or why they're not unloaded yet. I'm still working through debugging the problem.

What I know so far:
Encountered only when blackbird is in use.
Metaspace  is filled over time.
Loaded classes as reported by the JVM approaches 2 million (where I'd expect tens of thousands)
Impacts java 15. Maybe others, but I don't have a reproducer to validate whether or not only java 15 is impacted.

stevenschlansker:
Fixes #147

### DIFF
--- a/blackbird/README.md
+++ b/blackbird/README.md
@@ -17,6 +17,13 @@ implementations, we can write needed adapters as simple Java code and use the me
 to create distinct call sites for every needed access path.  This should allow each accessor to
 have a monomorphic call profile and easily inline for maximum performance.
 
+> **Warning**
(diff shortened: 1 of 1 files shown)


cowtowncoder (approved):

cowtowncoder:
Thank you @stevenschlansker

Issue #191, opened 2023-01-08
graalvm support for blackbird


ashraf-revo:
iam getting this errors when using blackbird 
`java.lang.InternalError: com.oracle.svm.core.jdk.UnsupportedFeatureError: Defining hidden classes at runtime is not supported.
	at java.base@17.0.5/java.lang.invoke.InnerClassLambdaMetafactory.generateInnerClass(InnerClassLambdaMetafactory.java:413) ~[na:na]
	at java.base@17.0.5/java.lang.invoke.InnerClassLambdaMetafactory.spinInnerClass(InnerClassLambdaMetafactory.java:315) ~[na:na]
	at java.base@17.0.5/java.lang.invoke.InnerClassLambdaMetafactory.buildCallSite(InnerClassLambdaMetafactory.java:228) ~[na:na]
	at java.base@17.0.5/java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:341) ~[na:na]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.createProperty(BBSerializerModifier.java:155) ~[na:na]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.lambda$findProperties$0(BBSerializerModifier.java:67) ~[na:na]
	at com.fasterxml.jackson.module.blackbird.util.Unchecked.lambda$runnable$0(Unchecked.java:31) ~[na:na]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.findProperties(BBSerializerModifier.java:68) ~[na:na]
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.changeProperties(BBSerializerModifier.java:52) ~[na:na]
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.constructBeanOrAddOnSerializer(BeanSerializerFactory.java:415) ~[na:na]
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.findBeanOrAddOnSerializer(BeanSerializerFactory.java:295) ~[na:na]
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory._createSerializer2(BeanSerializerFactory.java:240) ~[na:na]
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.createSerializer(BeanSerializerFactory.java:174) ~[na:na]
[...]

PR #217, opened 2023-07-31
Follow up wrt #191 by using `NativeImageUtil.isRunningInNativeImage()` in `databind`

JooHyukKim:
### Motivation

Seems like the `TODO:` was written before exposing `NativeImageUtil.isRunningInNativeImage()` in `databind` by https://github.com/FasterXML/jackson-databind/pull/4060

### DIFF
--- a/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/AfterburnerModule.java
+++ b/afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/AfterburnerModule.java
@@ -2,14 +2,20 @@ package com.fasterxml.jackson.module.afterburner;
 
 import com.fasterxml.jackson.core.Version;
 import com.fasterxml.jackson.databind.Module;
+import com.fasterxml.jackson.databind.util.NativeImageUtil;
 import com.fasterxml.jackson.module.afterburner.ser.SerializerModifier;
 import com.fasterxml.jackson.module.afterburner.deser.DeserializerModifier;
 
 public class AfterburnerModule extends Module
     implements java.io.Serializable // is this necessary?
 {
-    // TODO: replace with jackson-databind/NativeImageUtil.RUNNING_IN_SVM
-    private static final boolean RUNNING_IN_SVM = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
+    /**
+     * <a href="https://github.com/FasterXML/jackson-modules-base/issues/191">[modules-base#191] Native image detection</a>
+     * 
+     * @since 2.16
+     */
+    private static final boolean RUNNING_IN_SVM = NativeImageUtil.isRunningInNativeImage();
+    
     private static final long serialVersionUID = 1L;
 
     /*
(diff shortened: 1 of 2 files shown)


afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/AfterburnerModule.java
```diff
@@ -2,14 +2,20 @@
...
     implements java.io.Serializable // is this necessary?
 {
-    // TODO: replace with jackson-databind/NativeImageUtil.RUNNING_IN_SVM
-    private static final boolean RUNNING_IN_SVM = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
+    /**
+     * <a href="https://github.com/FasterXML/jackson-modules-base/issues/191">[modules-base#191] Native image detection</a>
+     * 
+     * @since 2.16
+     */
+    private static final boolean RUNNING_IN_SVM = NativeImageUtil.isRunningInNativeImage();
```
cowtowncoder, line 17: Not sure this can be done, since method Javadoc states:

```
     * This check cannot be a constant, because
     * the static initializer may run early during build time
```

cowtowncoder: Yes, I think inlining would make sense: it will only be called once anyway.

cowtowncoder: Not quite sure why logic uses part-static, part-dynamic logic there. I assume it's just an optimization to avoid dynamic check in some cases... but it really does seem odd. I think @yawkat might be able to shed more light on this.

yawkat: as discussed in #216, `isRunningInNativeImage()` should not be used here. There needs to be a new getter for `RUNNING_IN_SVM` that can be used instead.

yawkat: maybe `isRunningInNativeImage()` should be renamed also. i did not think enough about the name when i wrote it, since it wasn't public api. now that it is, a better name would be appropriate.

cowtowncoder: Since the method was non-public before 2.16, it's fine to change the name I think.

afterburner/src/main/java/com/fasterxml/jackson/module/afterburner/AfterburnerModule.java
```diff
@@ -51,7 +50,8 @@ public AfterburnerModule() { }
     @Override
     public void setupModule(SetupContext context)
     {
-        if (RUNNING_IN_SVM)
+        // [modules-base#191] Since 2.16, Native image detection 
+        if (NativeImageUtil.isInSVM())
```
stevenschlansker, line 54: I think you missed updating this line with the new name.

stevenschlansker (approved):

PR #227, opened 2023-10-27
Add Android Record Module

eranl:
Add a module that allows deserialization into records using the canonical constructor on Android, where java records are supported through desugaring, and Jackson's built-in support for records doesn't work, since the desugared classes have a non-standard super class, and record component-related reflection methods are missing.

See [Android Developers Blog article](https://android-developers.googleblog.com/2023/06/records-in-android-studio-flamingo.html)

Note: the canonical record constructor is found through matching of parameter types with field types. Therefore, this module doesn't allow a deserialized desugared record class to have a custom
constructor with a signature that's any permutation of the canonical one's.

See https://github.com/FasterXML/jackson-databind/issues/4166

This is the simplest implementation I could make. Please let me know if I should handle more cases.


android-record/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+        <!-- Generate PackageVersion.java into this directory. -->
+        <packageVersion.dir>com/fasterxml/jackson/module/androidrecord</packageVersion.dir>
+        <packageVersion.package>com.fasterxml.jackson.module.androidrecord</packageVersion.package>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
```
pjfanning, line 39: Could you add jackson-core as a dependency? It is better not to rely on it transitively when you have code that references jackson-core classes.

pjfanning (commented):
Could you use the animal sniffer maven plugin to ensure that we support a specific Android SDK version? It would be useful to set the android version to the lowest one that works - so that we can document the minimum supported Android SDK version. 

See https://github.com/FasterXML/jackson-core/blob/2.16/pom.xml - jackson-core tests that we support Android SDK version 26.

https://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/

android-record/pom.xml
eranl: Done. Should I add a dependency on jackson-annotations as well?

eranl:
> Could you use the animal sniffer maven plugin to ensure that we support a specific Android SDK version? It would be useful to set the android version to the lowest one that works - so that we can document the minimum supported Android SDK version.
> 
> See https://github.com/FasterXML/jackson-core/blob/2.16/pom.xml - jackson-core tests that we support Android SDK version 26.
> 
> https://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/

Done. I put the version settings in the main POM file. Is that correct?
Also, there's no point in using a lower Android SDK version than 26, right?

android-record/pom.xml
cowtowncoder: If there is direct dependency to one or more annotations classes, yes.

android-record/src/moditect/module-info.java
```diff
@@ -0,0 +1,9 @@
+module com.fasterxml.jackson.module.androidrecord {
+
+    requires com.fasterxml.jackson.databind;
```
cowtowncoder, line 3: Jackson core and annotations here as well, I think.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        SettableBeanProperty[] properties = new SettableBeanProperty[parameters.length];
+        for (int i = 0; i < parameters.length; i++) {
+          Parameter parameter = parameters[i];
+          properties[i] = CreatorProperty.construct(PropertyName.construct(parameter.getName()), config.getTypeFactory()
+                          .constructType(parameter.getParameterizedType()), null, null, null, null, i, null, null);
+        }
+
+        ((StdValueInstantiator) defaultInstantiator).configureFromObjectSettings(null, null, null, null,
```
cowtowncoder, line 82: This looks dangerous... having mostly `null`s, and overwriting internal state.
Ideally I'd rather create a new instance (even if of type `StdValueInstantiator`) instead of forcibly trying to modify passed-in instantiator

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private static ValueInstantiator findValueInstantiator(DeserializationConfig config, BeanDescription beanDesc,
+                                                         ValueInstantiator defaultInstantiator) {
+    Class<?> raw = beanDesc.getType().getRawClass();
+    if (defaultInstantiator instanceof StdValueInstantiator && raw.getSuperclass() != null
+            && raw.getSuperclass().getName().equals("com.android.tools.r8.RecordTag")) {
+      Map<Type, Integer> componentTypes = typeMap(Arrays.stream(raw.getDeclaredFields())
+              .filter(field -> !Modifier.isStatic(field.getModifiers())).map(Field::getGenericType));
+      boolean found = false;
+      for (Constructor<?> constructor: raw.getDeclaredConstructors()) {
```
cowtowncoder, line 63: For constructors it'd be better to access `AnnotatedConstructor`s from `beanDesc` if at all accessible.
Same goes for fields (-> `AnnotatedField`).
These would give you instances needed for constructing other objects, as well as processed annotations (including possible mix-ins).

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
```diff
@@ -0,0 +1,95 @@
...
+        SettableBeanProperty[] properties = new SettableBeanProperty[parameters.length];
+        for (int i = 0; i < parameters.length; i++) {
+          Parameter parameter = parameters[i];
+          properties[i] = CreatorProperty.construct(PropertyName.construct(parameter.getName()), config.getTypeFactory()
+                          .constructType(parameter.getParameterizedType()), null, null, null, null, i, null, null);
+        }
+
+        ((StdValueInstantiator) defaultInstantiator).configureFromObjectSettings(null, null, null, null,
+                           new AnnotatedConstructor(null, constructor, null, null), properties);
+        constructor.setAccessible(true);
```
cowtowncoder, line 84: There is a configuration setting that determines whether these calls should be made (`MapperFeature.CAN_OVERRIDE_ACCESS_MODIFIERS`).
And `ClassUtil` has method typically used, `checkAndFixAccess`, when forcing is enabled.

android-record/pom.xml
eranl: Done as well.
Actually, the module itself depends directly on jackson-databind and jackson-annotations only. Test code depends on all three. So should I mark the jackson-core dependency "<scope>test</scope>"?

android-record/src/moditect/module-info.java
eranl: Done. But see [question above](https://github.com/FasterXML/jackson-modules-base/pull/227#discussion_r1380933330).

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
eranl: I've improved the Property construction code. Please take a look.
Regarding modifying the instantiator, I followed [this example](https://github.com/FasterXML/jackson-datatype-jsr310/blob/master/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.java#L232). Should I use a different pattern?

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
eranl: > For constructors it'd be better to access `AnnotatedConstructor`s from `beanDesc` if at all accessible. 

Done, thanks.

> Same goes for fields (-> `AnnotatedField`). 

Couldn't find that.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
eranl: Done

eranl:
With the goal of maximizing consistency with built-in record support, I copied and "desugared" some unit tests from [jackson-databind](https://github.com/FasterXML/jackson-databind/tree/2.16/src/test-jdk17/java/com/fasterxml/jackson/databind/records). A few of the test cases are failing, and I marked them with a "Failing" comment and a "notest" name prefix. I'm hoping for guidance about whether and how I should fix them.
Also, should I copy all of the tests from that package?

android-record/pom.xml
cowtowncoder: Hmmh. No, I think either only include jackson-databind and jackson-annotations deps, or all 3, but not in "test" scope. Either would be fine.

android-record/src/moditect/module-info.java
cowtowncoder: Ok. I think databind+annotations or databind+core+annotations both fine, just prefer uses same set in pom and module-info.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
cowtowncoder: Will do. I think that usage pattern is probably fine -- will add a comment if I remember reasons to use different approach.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
cowtowncoder: Ah. Should have looked at it: method to look for is

    public abstract AnnotatedClass getClassInfo();

and `AnnotatedClass` has method `fields()` to get them.

cowtowncoder:
> With the goal of maximizing consistency with built-in record support, I copied and "desugared" some unit tests from [jackson-databind](https://github.com/FasterXML/jackson-databind/tree/2.16/src/test-jdk17/java/com/fasterxml/jackson/databind/records). A few of the test cases are failing, and I marked them with a "Failing" comment and a "notest" name prefix. I'm hoping for guidance about whether and how I should fix them. Also, should I copy all of the tests from that package?

That sounds good. I don't think you necessarily need to copy all tests, although I guess number of tests is not huge in grand scheme of things. It's just that there's then extra maintenance needed.

cowtowncoder:
Ok I will try to go over code again; I like how things are developing and my only concern is wrt 2.16.0 release whether we can make this ready to go in.

But in the meantime one thing that I will need to merge is CLA (assuming you haven't sent one earlier @eranl ?).
This is only needed once for any and all Jackson contributions; it's from here:

and the usual way is to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once this is done PR may be merged, assuming everything else is ready.

Looking forward to merging this & once again thank you for contributing it @eranl !

android-record/pom.xml
eranl: Ok, I'll leave all three.

android-record/src/moditect/module-info.java
eranl: Ok, done.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
eranl: Thanks. Unfortunately, it seems to not be useful here. This module uses three pieces of metadata about fields: name, modifiers, and generic type. AnnotatedField is a pass-through for the former two, and the latter is unavailable on it AFAICT.

eranl:
> > With the goal of maximizing consistency with built-in record support, I copied and "desugared" some unit tests from [jackson-databind](https://github.com/FasterXML/jackson-databind/tree/2.16/src/test-jdk17/java/com/fasterxml/jackson/databind/records). A few of the test cases are failing, and I marked them with a "Failing" comment and a "notest" name prefix. I'm hoping for guidance about whether and how I should fix them. Also, should I copy all of the tests from that package?
> 
> That sounds good. I don't think you necessarily need to copy all tests, although I guess number of tests is not huge in grand scheme of things. It's just that there's then extra maintenance needed.

What should I do about the failing tests? Fix them (I'd need guidance on that)? Delete them? Move them to a `failing` package?

eranl:
> > Could you use the animal sniffer maven plugin to ensure that we support a specific Android SDK version? It would be useful to set the android version to the lowest one that works - so that we can document the minimum supported Android SDK version.
> > See https://github.com/FasterXML/jackson-core/blob/2.16/pom.xml - jackson-core tests that we support Android SDK version 26.
> > https://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/
> 
> Done. I put the version settings in the main POM file. Is that correct? Also, there's no point in using a lower Android SDK version than 26, right?

Is this ok?

android-record/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+
+    <build>
+        <plugins>
+          <plugin>
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-compiler-plugin</artifactId>
+               <inherited>true</inherited>
+               <configuration>
+                 <compilerArgs>
+                   <arg>-parameters</arg>
```
cowtowncoder, line 58: Forgot to ask: is this necessary? Similarly, other plugin configuration? If so, maybe add XML comment to explain.

android-record/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+
+            <plugin>
+                <groupId>org.moditect</groupId>
+                <artifactId>moditect-maven-plugin</artifactId>
+            </plugin>
+
+            <plugin>
+              <groupId>org.codehaus.mojo</groupId>
+              <artifactId>animal-sniffer-maven-plugin</artifactId>
+              <version>${version.plugin.animal-sniffer}</version>
```
cowtowncoder, line 85: I guess this must come from `jackson-core` dependency?

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
```diff
@@ -0,0 +1,167 @@
...
+
+      return super.collectProperties(config, type, r, forSerialization);
+    }
+  }
+
+  @Override
+  public void setupModule(SetupContext context) {
+    super.setupModule(context);
+    context.addValueInstantiators(AndroidRecordModule::findValueInstantiator);
+    context.setClassIntrospector(new AndroidRecordClassIntrospector());
```
cowtowncoder, line 112: Hmmmmh. This is somewhat problematic as it does not compose; ideally Modules wouldn't have to override it, as only one module can really change it (others would override it).
I'll need to think about this a bit.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
```diff
@@ -0,0 +1,167 @@
...
+    return raw.getSuperclass() != null && raw.getSuperclass().getName().equals("com.android.tools.r8.RecordTag");
+  }
+
+  private static ValueInstantiator findValueInstantiator(DeserializationConfig config, BeanDescription beanDesc,
+                                                         ValueInstantiator defaultInstantiator) {
+    Class<?> raw = beanDesc.getType().getRawClass();
+    if (! defaultInstantiator.canCreateFromObjectWith() && defaultInstantiator instanceof StdValueInstantiator
+            && isDesugaredRecordClass(raw)) {
+      Map<String, Type> components = getDesugaredRecordComponents(raw)
+              .collect(Collectors.toMap(Field::getName, Field::getGenericType));
```
cowtowncoder, line 125: Instead of `getGenericType()`, just use (type-erased) `getType()` -- matching is done using type-erased types.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
```diff
@@ -0,0 +1,167 @@
...
+      boolean found = false;
+      for (AnnotatedConstructor constructor: beanDesc.getConstructors()) {
+        Parameter[] parameters = constructor.getAnnotated().getParameters();
+        Map<String, Type> parameterTypes = Arrays.stream(parameters)
+                .collect(Collectors.toMap(Parameter::getName, Parameter::getParameterizedType));
```
cowtowncoder, line 130: as per above, Parameter::getType works as well.

cowtowncoder:
@eranl I think that due to timing constraints -- I really need to get this merged ASAP, if it is to make it in 2.16.0! -- I am ok with the implementation as is; we can tackle some changes in 2.17.
And as I said, implementation is mostly fine anyway.

But one thing that must occur before I can merge this is the CLA (see my earlier comment).
If at all possible, it'd be great if you could get that sent so I can proceed with merging.

As to failing tests; yes, moving them under `src/test/java/.../failing` would make sense (if not already moved), until issues can be resolved.
It may also make sense to trim `BaseTest` and `BaseMapTest` to only include methods actually needed. But that's icing on the cake.
One challenge is that I need to merge 2.x branch into 3.0 (master); that's some effort although mostly mechanical (different Java package, Maven artifact, module name).

android-record/pom.xml
eranl: It comes from the project's main pom file. See [discussion above](https://github.com/FasterXML/jackson-modules-base/pull/227#issuecomment-1783636041).

android-record/pom.xml
eranl: Added comment.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
eranl: ditto

eranl:
> But one thing that must occur before I can merge this is the CLA (see my earlier comment). If at all possible, it'd be great if you could get that sent so I can proceed with merging.

Sent.

> As to failing tests; yes, moving them under `src/test/java/.../failing` would make sense (if not already moved), until issues can be resolved. 

Done.

> It may also make sense to trim `BaseTest` and `BaseMapTest` to only include methods actually needed. But that's icing on the cake. 

Done.

android-record/pom.xml
cowtowncoder: Oh, I missed that one (main-level pom at the bottom of diffs). Thanks!

android-record/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <inherited>true</inherited>
+        <configuration>
+          <compilerArgs>
+            <!-- Module uses constructor parameter names (and types) to identify the canonical constructor -->
```
cowtowncoder, line 58: Ah. Because unlike Records there's no specified metadata to fix ordering, so names need to be used.

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
cowtowncoder: Hmmh. How is this possible? All matching in class files typically uses erased types, and cannot really use type parameters to resolve conflicts.

Unless, I guess, this is combined with parameter ordering differing. That's probably it and makes sense. So since Constructor parameter ordering not fixed wrt fields, can have same type erased set in different order.

cowtowncoder:
@eranl My apologies: I don't think we have `cla` at fasterxml dot com alias -- I'll try to set up one, but in the meantime, could you re-send CLA to `info` at fasterxml dot com instead.
Apologies for the mix up here.

android-record/pom.xml
eranl: Right. I've been wondering if I should document this requirement, but based on my local testing, it seems that the android build system always stores parameter names for desugared record canonical constructors in class files, at least with SDK 34.

eranl:
> @eranl My apologies: I don't think we have `cla` at fasterxml dot com alias -- I'll try to set up one, but in the meantime, could you re-send CLA to `info` at fasterxml dot com instead. Apologies for the mix up here.

Sent.

cowtowncoder (approved):

android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
eranl: Does that mean that my statement in the doc that "this module is a no-op when no Android-desugared records are being (de)serialized" is not accurate anymore? That *is* problematic.

cowtowncoder: Only if some other module tries to override `ClassIntrospector` -- it is not problem if this is the only module to do so.
So unlike, say, `AnnotationIntrospector`s that can be chained, there's no existing mechanism to combine `ClassIntrospector`s

eranl: How common is it for modules to do that? Should I document this?

cowtowncoder: @eranl It might be best to add something in README to mention this.
I don't remember any other module overriding it, so this is theoretical problem, not actual one.
(at least wrt modules under FasterXML)

On chaining... I thought about that, but in general I'm not sure how commonly this is needed etc.
Due to timing it needs to wait until 2.17 (that is, post 2.16.0 and cannot go in a patch) so there's no hurry.
So what I am thinking is that for 2.17 we could consider various approaches to remove this problem; and not have to make last minute changes that may turn out problematic.

eranl: > @eranl It might be best to add something in README to mention this. I don't remember any other module overriding it, so this is theoretical problem, not actual one. (at least wrt modules under FasterXML)

Done. see #230.

> On chaining... I thought about that, but in general I'm not sure how commonly this is needed etc. Due to timing it needs to wait until 2.17 (that is, post 2.16.0 and cannot go in a patch) so there's no hurry. So what I am thinking is that for 2.17 we could consider various approaches to remove this problem; and not have to make last minute changes that may turn out problematic.

Sure, I meant in a future version. Please let me know when is a good time, and what's your preferred approach.

android-record/pom.xml
cowtowncoder: Thank you @eranl --merged!

PR #229, opened 2023-11-11
feat: add subtype module

black-06:
Adds a module that allows subtype to be registered without annotating the parent class.

It is implemented on SPI.

See https://github.com/FasterXML/jackson-databind/issues/2104

Its original version is https://github.com/black-06/jackson-modules-dynamic-subtype.

I removed the custom ServiceLoader and always use the standard library,
which means that pojo always needs a no-parameter constructor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,7 @@ not datatype, data format, or JAX-RS provider modules.
     <module>mrbean</module>
     <module>osgi</module>
     <module>paranamer</module>
+    <module>subtype</module>
     <!-- since 2.13: -->
     <module>no-ctor-deser</module>
   </modules>
--- a/subtype/README.md
+++ b/subtype/README.md
@@ -1,0 +1,66 @@
+# jackson-module-subtype
+
+Registering subtypes without annotating the parent class,
+see [this](https://github.com/FasterXML/jackson-databind/issues/2104).
+
+Implementation on SPI.
+
+# Usage
+
+Registering modules.
+
+```
+ObjectMapper mapper = new ObjectMapper().registerModule(new DynamicSubtypeModule());
+```
+
+Ensure that the parent class has at least the `JsonTypeInfo` annotation.
+
+```java
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public interface Parent {
+}
+```
+
+1. add the `JsonSubType` annotation to your subclass.
+2. provide a non-argument constructor (SPI require it).
+
+```java
+import io.github.black.jackson.JsonSubType;
+
+@JsonSubType("first-child")
+public class FirstChild {
+
+    private String foo;
+    // ...
+
+    public FirstChild() {
+    }
+}
+```
+
+SPI: Put the subclasses in the `META-INF/services` directory under the interface.
+Example: `META-INF/services/package.Parent`
+
+```
+package.FirstChild
+```
+
+Alternatively, you can also use the `auto-service` to auto-generate these files:
+
+```java
+import io.github.black.jackson.JsonSubType;
+import com.google.auto.service.AutoService;
+
+@AutoService(Parent.class)
+@JsonSubType("first-child")
+public class FirstChild {
+
+    private String foo;
+    // ...
+
+    public FirstChild() {
+    }
+}
+```
+
+Done, enjoy it.
--- a/subtype/pom.xml
+++ b/subtype/pom.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <!-- This module was also published with a richer model, Gradle metadata,  -->
+    <!-- which should be used instead. Do not delete the following line which  -->
+    <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
+    <!-- that they should prefer consuming it instead. -->
+    <!-- do_not_remove: published-with-gradle-metadata -->
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.fasterxml.jackson.module</groupId>
+        <artifactId>jackson-modules-base</artifactId>
+        <version>2.16.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>jackson-module-subtype</artifactId>
+    <name>Jackson module: Subtype Annotation Support</name>
+    <packaging>bundle</packaging>
+
+    <description>Registering subtypes without annotating the parent class</description>
+    <url>https://github.com/FasterXML/jackson-modules-base</url>
+
+    <licenses>
+        <license>
+            <name>The Apache Software License, Version 2.0</name>
+            <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <properties>
+        <!-- Generate PackageVersion.java into this directory. -->
+        <packageVersion.dir>com/fasterxml/jackson/module/subtype</packageVersion.dir>
+        <packageVersion.package>com.fasterxml.jackson.module.subtype</packageVersion.package>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+            <version>1.0.1</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.google.code.maven-replacer-plugin</groupId>
+                <artifactId>replacer</artifactId>
+            </plugin>
+            <!--  14-Mar-2019, tatu: Add rudimentary JDK9+ module info. To build with JDK 8
+              will have to use `moduleInfoFile` as anything else requires JDK 9+
+              -->
+            <plugin>
+                <groupId>org.moditect</groupId>
+                <artifactId>moditect-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>9</source>
+                    <target>9</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/subtype/src/main/java/com/fasterxml/jackson/module/subtype/JsonSubType.java
+++ b/subtype/src/main/java/com/fasterxml/jackson/module/subtype/JsonSubType.java
@@ -1,0 +1,41 @@
+package com.fasterxml.jackson.module.subtype;
+
+import com.fasterxml.jackson.annotation.JacksonAnnotation;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Definition of a subtype, along with optional name(s). If no name is defined
+ * (empty Strings are ignored), class of the type will be checked for {@link JsonTypeName}
+ * annotation; and if that is also missing or empty, a default
+ * name will be constructed by type id mechanism.
+ * Default name is usually based on class name.
+ * <p>
+ * It's the same as {@link  JsonSubTypes.Type}.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@JacksonAnnotation
+public @interface JsonSubType {
+    /**
+     * Logical type name used as the type identifier for the class, if defined; empty
+     * String means "not defined". Used unless {@link #names} is defined as non-empty.
+     *
+     * @return subtype name
+     */
+    String value() default "";
+
+    /**
+     * (optional) Logical type names used as the type identifier for the class: used if
+     * more than one type name should be associated with the same type.
+     *
+     * @return subtype name array
+     * @since 2.12
+     */
+    String[] names() default {};
+}
--- a/subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
+++ b/subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
@@ -1,0 +1,120 @@
+package com.fasterxml.jackson.module.subtype;
+
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.Module;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.jsontype.NamedType;
+import com.fasterxml.jackson.module.subtype.PackageVersion;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Subtype module.
+ * <p>
+ * The module caches the subclass, so it's non-real-time.
+ * It's for registering subtypes without annotating the parent class.
+ * See <a href="https://github.com/FasterXML/jackson-databind/issues/2104">this issues</a> in jackson-databind.
+ * <p>
+ * When not found in the cache, it loads and caches subclasses using SPI.
+ * Therefore, we can {@link #unregisterType} a class and then module will reload this class's subclasses.
+ */
+public class SubtypeModule extends Module {
+
+    private final ConcurrentHashMap<Class<?>, List<NamedType>> subtypes = new ConcurrentHashMap<>();
+
+    @Override
+    public String getModuleName() {
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public Version version() {
+        return PackageVersion.VERSION;
+    }
+
+    @Override
+    public void setupModule(SetupContext context) {
+        context.insertAnnotationIntrospector(new AnnotationIntrospector() {
+            @Override
+            public Version version() {
+                return PackageVersion.VERSION;
+            }
+
+            @Override
+            public List<NamedType> findSubtypes(Annotated a) {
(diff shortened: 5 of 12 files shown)


pom.xml
```diff
@@ -33,6 +33,7 @@ not datatype, data format, or JAX-RS provider modules.
     <module>mrbean</module>
     <module>osgi</module>
     <module>paranamer</module>
+    <module>subtype</module>
```
cowtowncoder, line 36: I'll probably want to change the name, "subtype" is too generic. But let me think about better name for a while...

subtype/README.md
```diff
@@ -0,0 +1,66 @@
...
+see [this](https://github.com/FasterXML/jackson-databind/issues/2104).
+
+Implementation on SPI.
+
+# Usage
+
+Registering modules.
+
+```
+ObjectMapper mapper = new ObjectMapper().registerModule(new DynamicSubtypeModule());
```
cowtowncoder, line 13: I think current name is `SubtypeModule`.

subtype/README.md
```diff
@@ -0,0 +1,66 @@
...
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public interface Parent {
+}
+```
+
+1. add the `JsonSubType` annotation to your subclass.
+2. provide a non-argument constructor (SPI require it).
+
+```java
+import io.github.black.jackson.JsonSubType;
```
cowtowncoder, line 28: Needs to change to new package.

subtype/src/test/java/com/fasterxml/jackson/module/subtype/WithJsonSubTypesTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * test work with {@link JsonSubTypes}
+ */
+@RunWith(value = Parameterized.class)
+public class WithJsonSubTypesTest<T extends WithJsonSubTypesTest.Parent> {
+
+    private final ObjectMapper mapper = new ObjectMapper().findAndRegisterModules();
```
cowtowncoder, line 26: Let's not use this method but instead explicitly register module.

subtype/src/test/java/com/fasterxml/jackson/module/subtype/WithJsonSubTypesTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+        @SuppressWarnings("unused") // SPI require it
+        public FirstChild() {
+        }
+
+        public FirstChild(String foo) {
+            this.foo = foo;
+        }
+
+        @SuppressWarnings("unused") // jackson require it
```
cowtowncoder, line 85: We could instead just use `public String foo` to avoid having to add getter and setter

subtype/src/test/java/com/fasterxml/jackson/module/subtype/WithoutJsonSubTypesTest.java
```diff
@@ -0,0 +1,121 @@
...
+import com.google.auto.service.AutoService;
+import org.junit.Test;
+
+import java.util.Objects;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * test work without {@link JsonSubTypes}
```
cowtowncoder, line 15: I am not sure I understand -- why would we test functioning without module? And isn't module actually register with the mapper. Or what is being tested here? Could probably use longer explanation of reasoning.

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,120 @@
...
+            public Version version() {
+                return PackageVersion.VERSION;
+            }
+
+            @Override
+            public List<NamedType> findSubtypes(Annotated a) {
+                registerTypes(a.getRawType());
+
+                List<NamedType> list1 = SubtypeModule.findSubtypes(a.getRawType(), a::getAnnotation);
+                List<NamedType> list2 = subtypes.getOrDefault(a.getRawType(), Collections.emptyList());
```
JooHyukKim, line 54: Can we rename these `list1` and `list2` to something less generic?

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,120 @@
...
+
+                if (list1.isEmpty()) return list2;
+                if (list2.isEmpty()) return list1;
```
JooHyukKim, line 57: If the two lists are null-safe, no need for empty checking.

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,120 @@
...
+                List<NamedType> list = new ArrayList<>(list1.size() + list2.size());
+                list.addAll(list1);
+                list.addAll(list2);
+                return list;
```
JooHyukKim, line 61:
```suggestion
                List<NamedType> list = new ArrayList<>(list1.size() + list2.size());
                list1.addAll(list2);
                return list;
```

JooHyukKim, line 61: Is it possible to have duplication? If so, can we have these as set first, then return as list?

subtype/src/test/java/com/fasterxml/jackson/module/subtype/WithoutJsonSubTypesTest.java
JooHyukKim: Probably test against Jackson polymorphic type handling working as expected. If so, does not seem necessary

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,120 @@
...
+            result.addAll(findSubtypes(subclass, subclass::getAnnotation));
+        }
+        subtypes.put(parent, result);
+    }
+
+    public void unregisterType(Class<?> parent) {
+        subtypes.remove(parent);
+    }
+
+    private static <S> List<NamedType> findSubtypes(Class<S> clazz, Function<Class<JsonSubType>, JsonSubType> getter) {
```
JooHyukKim, line 104: Is there a reason for keeping it static? if not maybe like...
```suggestion
    private <S> List<NamedType> _findSubtypes(Class<S> clazz, Function<Class<JsonSubType>, JsonSubType> getter) {
```

subtype/src/main/java/com/fasterxml/jackson/module/subtype/JsonSubType.java
```diff
@@ -0,0 +1,41 @@
...
+ * annotation; and if that is also missing or empty, a default
+ * name will be constructed by type id mechanism.
+ * Default name is usually based on class name.
+ * <p>
+ * It's the same as {@link  JsonSubTypes.Type}.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@JacksonAnnotation
+public @interface JsonSubType {
```
JooHyukKim, line 24: Name sounds like it's part of `@JsonSubTypes` of `annotations` module. This may confuse users quite. If module name changes as per https://github.com/FasterXML/jackson-modules-base/pull/229/files#r1390329649, can we change name here also?

subtype/src/main/java/com/fasterxml/jackson/module/subtype/JsonSubType.java
```diff
@@ -0,0 +1,41 @@
...
+     * @return subtype name
+     */
+    String value() default "";
+
+    /**
+     * (optional) Logical type names used as the type identifier for the class: used if
+     * more than one type name should be associated with the same type.
+     *
+     * @return subtype name array
+     * @since 2.12
```
JooHyukKim, line 38: Version mismatch, can remove.
```suggestion
```

... then add `@since 2.16` to the class

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,120 @@
...
+ * Subtype module.
+ * <p>
+ * The module caches the subclass, so it's non-real-time.
+ * It's for registering subtypes without annotating the parent class.
+ * See <a href="https://github.com/FasterXML/jackson-databind/issues/2104">this issues</a> in jackson-databind.
+ * <p>
+ * When not found in the cache, it loads and caches subclasses using SPI.
+ * Therefore, we can {@link #unregisterType} a class and then module will reload this class's subclasses.
+ */
+public class SubtypeModule extends Module {
```
JooHyukKim, line 27:
```suggestion
 * Therefore, we can {@link #unregisterType} a class and then module will reload this class's subclasses.
 *
 * @since 2.16
 */
public class SubtypeModule extends Module {
```

pom.xml
black-06: It's a difficult thing for me to give it a name...

subtype/src/test/java/com/fasterxml/jackson/module/subtype/WithoutJsonSubTypesTest.java
black-06: I actually wanted to test whether `@JsonSubType` worked with `@JsonSubTypes`

pom.xml
JooHyukKim: True🥲🥲

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
black-06: Here is a simple modification of https://github.com/FasterXML/jackson-databind/blob/3a271be870821301f73054e43e32f657e4753234/src/main/java/com/fasterxml/jackson/databind/introspect/AnnotationIntrospectorPair.java#L280-L291

cowtowncoder:
Quick note: due to timing, decided that I will not try to get this in Jackson 2.16 -- so will need to be re-based to 2.17.
This because there's a bit of work to be done here.

For example, I think that naming of `@JsonSubType` should change as the general rule is that only `jackson-annotations` has annotations with name starting with `@Json` (and for legacy reasons, some in `jackson-databind`).
Modules can provide annotations that start with `@Jackson` prefix (or anything else that makes sense).

But I need to read the PR in bit more thought to make sure my feedback is relevant.

amricko0b:
Hello everyone!
Any updates on this one?

cowtowncoder:
I totally forgot. If this was to be added, would need to re-base/re-create against 2.x branch.

black-06:
> I totally forgot. If this was to be added, would need to re-base/re-create against 2.x branch.

I've changed base to 2.x branch, and I checked the comments above, most of them have been modified, and the problems that haven't been solved include:
- findSubtypes: It's almost a copy of [AnnotationIntrospectorPair.findSubtypes](https://github.com/FasterXML/jackson-databind/blob/2.x/src/main/java/com/fasterxml/jackson/databind/introspect/AnnotationIntrospectorPair.java#L251), so I tend not to modify it.
- annotation name `JsonSubType`: If the user understands `JsonSubTypes`, then he should be able to understand `JsonSubType` easily, but if there is a more appropriate name, I have no problem.

Because it's been a while. We can review the code again.

pom.xml
cowtowncoder: ... we still need alternative name. :)

subtype/src/main/java/com/fasterxml/jackson/module/subtype/JsonSubType.java
cowtowncoder: Should be renamed, yes. Too confusing otherwise.

I think it should use `@Jackson` prefix, not `@Json`, for one. Maybe `@JacksonSubTypeName`

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,122 @@
...
+ * See <a href="https://github.com/FasterXML/jackson-databind/issues/2104">this issues</a> in jackson-databind.
+ * <p>
+ * When not found in the cache, it loads and caches subclasses using SPI.
+ * Therefore, we can {@link #unregisterType} a class and then module will reload this class's subclasses.
+ *
+ * @since 2.16
+ */
+public class SubtypeModule extends Module {
+
+    private final ConcurrentHashMap<Class<?>, List<NamedType>> subtypes = new ConcurrentHashMap<>();
```
cowtowncoder: Modules should be stateless; need to move in `AnnotationIntrospector` implementation instead (just refactoring).

May also lead to unexpected Class retention but that's... probably something to document, hard to avoid (without complexities of Soft/Weak ref Maps).

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeModule.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public Version version() {
+        return PackageVersion.VERSION;
+    }
+
+    @Override
+    public void setupModule(SetupContext context) {
+        context.insertAnnotationIntrospector(new AnnotationIntrospector() {
```
cowtowncoder: I think this `AnnotationIntrospector` should be just extracted into its own class.

subtype/src/main/resources/META-INF/LICENSE
```diff
@@ -0,0 +1,8 @@
+This copy of Jackson JSON processor `jackson-module-guice` module is licensed under the
```
cowtowncoder: guice -> module name

cowtowncoder:
This seems to fail on Java 8: if we are targeting Jackson 2.x, module should be Java 8 compatible.

But looks like simple change to `pom.xml` solves that. I'll fix it.

**EDIT**: done

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeAnnotationIntrospector.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * load parent's subclass by SPI.
+     *
+     * @param parent parent class.
+     * @param <S>    parent class type.
+     */
+    @SuppressWarnings("unchecked")
+    public <S> void registerTypes(Class<S> parent) {
```
cowtowncoder: Why is this `public`? Instead of package private instead?

subtype/src/main/java/com/fasterxml/jackson/module/subtype/SubtypeAnnotationIntrospector.java
```diff
@@ -0,0 +1,116 @@
...
+        subtypes.put(parent, result);
+    }
+
+    /**
+     * remove the parent class in the cache,
+     * so that {@link #registerTypes(Class)} can re-look by SPI.
+     *
+     * @param parent: parent class.
+     */
+    public void unregisterType(Class<?> parent) {
```
cowtowncoder: Why is this exposed?

subtype/src/main/java/com/fasterxml/jackson/module/subtype/JsonSubType.java
cowtowncoder: Let's go with `@JacksonSubType` (second alternative would be `@JacksonSpiSubType` but I'm ok leaving "SPI" here).
Only `jackson-annotations` should have `@Json`-prefixed annotations; not modules.

pom.xml
cowtowncoder: let's name directory "spi-subtypes".

subtype/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+    <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
+    <!-- that they should prefer consuming it instead. -->
+    <!-- do_not_remove: published-with-gradle-metadata -->
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.fasterxml.jackson.module</groupId>
+        <artifactId>jackson-modules-base</artifactId>
+        <version>2.21.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>jackson-module-subtype</artifactId>
```
cowtowncoder: Although we can leave "SPI" out from most places, I think module should be

    jackson-module-spi-subtypes

(adding `spi-` and plural `-s`)

cowtowncoder:
@black-06 Ok: I have some minor questions, but overall the only changes I do require are to naming:

1. Annotation name to avoid "@Json" prefix: my proposal is "@JacksonSubType"
2. Module name: `jackson-module-spi-subtypes` (not just "jackson-module-subtype")
3. Module directory: `spi-subtypes` (not just "subtypes")

and matching changes to `module-info.java` and README (aside from obvious Java classes and `pom.xml`).
I am open to other names too, but want them to be slightly more specific.

Once naming is changed, I can get this merge in 2.x for inclusion in upcoming 2.21.0 release.

cowtowncoder:
@black-06 One other (hopefully last) thing: if you haven't yet sent CLA (I don't think so but just in case), we would need it from:

https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf

This only needs to be done once, before merging the first PR: one is valid for all future contributions.
The usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once we have it, merging can occur (assuming code review etc complete).

Looking forward to getting this PR merged; thank you again for providing it!

subtype/pom.xml
black-06: > `registerTypes` Why is this public? Instead of package private instead?
> `unregisterType` Why is this exposed?

My previous idea was that the module allows SPI to register subtypes automatically, and also allows users to register/unregister subtypes manually.
Therefore, `registerTypes` is public, and `unregisterType` is provided (which is not called anywhere), and the module name does not have "SPI", because it can be set manually by users without using SPI.

If you think this is unnecessary, then the module name can be changed to `jackson-module-spi-subtypes`, which focuses on SPI registering subtypes by SPI.

cowtowncoder: Users can already register subtypes without this module (via `ObjectMapper.registerSubtypes()` directly or using `SimpleModule`). So yes, I think SPI is the interesting part.

Let's do renaming so I can merge this in for inclusion in 2.21 release.

black-06:
> @black-06 Ok: I have some minor questions, but overall the only changes I do require are to naming:
> 
> 1. Annotation name to avoid "@JSON" prefix: my proposal is "@JacksonSubType"
> 2. Module name: `jackson-module-spi-subtypes` (not just "jackson-module-subtype")
> 3. Module directory: `spi-subtypes` (not just "subtypes")
> 
> and matching changes to `module-info.java` and README (aside from obvious Java classes and `pom.xml`). I am open to other names too, but want them to be slightly more specific.
> 
> Once naming is changed, I can get this merge in 2.x for inclusion in upcoming 2.21.0 release.

The name has been changed. In addition, 
1. rename package to `com.fasterxml.jackson.module.spisubtypes`.
2. rename `SubtypeModule` to `SubtypesModule`, `SubtypeAnnotationIntrospector` to `SubtypesAnnotationIntrospector`.

Thank you very much for your review, ps. CLA has been sent.

cowtowncoder:
@black-06 Thank you for doing all of that. I am hoping to go over PR one more time, get it merged, and soon prepare for 2.21.0 release that includes the new module!

PR #235, opened 2024-01-12
Allow jackson-module-jaxb-annotations to tolerate 2.2,3 in Import-Package

### LINKED ISSUE #233: Tolerate JAX-RS 2.2 in jackson-module-jaxb-annotations so that it can be deployed in Liberty alongside features which use 2.2
When using a Liberty container image such as `icr.io/appcafe/websphere-liberty:23.0.0.9-full-java11-openj9-ubi` and trying to use the jackson-module-jaxb-annotations-2.16.0.jar we get a conflict due to two implementations of `javax.xml.bind` being available.

The features we need to use within Liberty only supply the 2.2 implementation.

Looking at [jackson-module-jaxb-annotations-2.16.0.jar](https://mvnrepository.com/artifact/com.fasterxml.jackson.module/jackson-module-jaxb-annotations/2.16.0) we can see it has compile dependencies on `jakarta.xml.bind-api` 2.3.2 and so provides an implementation at 2.3. This then causes the conflict.

This issue is to request that jaxb annotations be updated in some way to change the toleration that ends up in the /META-INF/MANIFEST.MF file in the Import-Package section to be:
```
javax.xml.bind;version="[2.2,3)",javax.xml.bind.annotation;
 version="[2.2,3)",javax.xml.bind.annotation.adapters;version="[2.2,3)
```
Instead of what is currently has:
```
javax.xml.bind;version="[2.3,3)",javax.xml.bind.annotation;
 version="[2.3,3)",javax.xml.bind.annotation.adapters;version="[2.3,3)
```

Thanks

JDUNNIN:
Resolves #233 

I removed the Jakarta import, from the comment `Despite groupId, this is "old" JAXB, not Jakarta (3.x is real Jakarta one)` it sounds like this should be OK. The end result in the MANIFEST.MF was the same as when I just added the jaxb-api dependency above it.

image

### DIFF
--- a/jaxb/pom.xml
+++ b/jaxb/pom.xml
@@ -54,14 +54,12 @@ for configuring data-binding.
       <artifactId>jackson-databind</artifactId>
     </dependency>
 
-    <!--  and actual JAXB annotations, types -->
-    <!-- NOTE! Despite groupId, this is "old" JAXB, not Jakarta
-         (3.x is real Jakarta one)
-      -->
+    <!-- Tolerate [2.2,3) for javax.xml.bind Import-Package in the generated MANIFEST.MF -->
+    <!-- https://mvnrepository.com/artifact/javax.xml.bind/jaxb-api -->
     <dependency>
-      <groupId>jakarta.xml.bind</groupId>
-      <artifactId>jakarta.xml.bind-api</artifactId>
-      <version>${version.jaxb.impl}</version>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+      <version>2.2.12</version>
     </dependency>
 
     <!-- 14-Mar-2019, tatu: Looks like this is needed for JDK11 and later


jaxb/pom.xml
```diff
@@ -54,14 +54,12 @@ for configuring data-binding.
...
-      -->
+    <!-- Tolerate [2.2,3) for javax.xml.bind Import-Package in the generated MANIFEST.MF -->
+    <!-- https://mvnrepository.com/artifact/javax.xml.bind/jaxb-api -->
     <dependency>
-      <groupId>jakarta.xml.bind</groupId>
-      <artifactId>jakarta.xml.bind-api</artifactId>
-      <version>${version.jaxb.impl}</version>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+      <version>2.2.12</version>
```
cowtowncoder, line 62: Shouldn't this instead change `version.jaxb.impl` property?

cowtowncoder:
Hmmh. I am really uncomfortable with this change -- I know JAXB/Jakarta upgrade has been a Royal Mess, and I am not sure this helps in general. It seems to me that there are TWO overlapping distinct packages (as in, different Maven coordinates) that nonetheless both provide same classes: one with `2.2.x` version (one chosen here) and another `2.3.x` (one being replaced). I don't think Maven can express such either-or case.
It seems like this could easily break existing usage with 2.3 dependency?

And in particular making change in a patch of 2.16, if breaking, would be very unfortunate. Especially as we are downgrading a dependency.

Is it not possible to add local overrides on package depending on JAXB annotations module?

@GedMarc WDYT?

JDUNNIN:
FYI we had a similar change go through in swagger-core https://github.com/swagger-api/swagger-core/pull/4595

cowtowncoder:
@JDUNNIN that is for OSGi dependnecy tho? Is the end results same?

One other quick note: since 2.17.1 is released, this PR needs to be rebased against 2.18; will not add in a patch.

JDUNNIN:
@cowtowncoder yes I had originally raised PR for fix the same as this PR, but they changed it to use the OSGi method instead. The end result was the same, MANIFEST.MF updated.

jaxb/pom.xml
JDUNNIN: No I don't think so as there is only a 2.3.3 or 2.3.2 available.

cowtowncoder: Looks like `version.jaxb.impl` is used to get a JAXB impl for tests:

```
    <!-- may also need JAXB impl for tests -->
    <dependency>
      <version>${version.jaxb.impl}</version>
      <scope>test</scope>
    </dependency>
```

cowtowncoder: ... but that does not seem to be needed, so I think I'll comment it out.

cowtowncoder (approved):

cowtowncoder:
UGGGGH. How did this end up merging 2.18 changes to 2.16 branch.... need to untangle this now.

cowtowncoder:
Ok had to hard reset 2.16 branch. What I do not understand is how Github managed to merge this into 2.16 branch, despite PR clearly stating 2.18 as target...

Need to re-create PR now.

PR #244, opened 2024-05-20
Re-apply #235 but now with 2.18 target (not sure what happened earlier)

cowtowncoder:


cowtowncoder:
@JDUNNIN Ok, merged again. Thank you for the PR!

JDUNNIN:
@cowtowncoder thank you for this! I assume I got something wrong with my other PR getting up to date with 2.18 but also don't understand how it went elsewhere even though the base was 2.18?!

cowtowncoder:
@JDUNNIN right. I do know that trying to change base of PR from higher (like 2.18) to lower (like 2.16) typically won't work like I want to (trying to merge 2.16..2.18 changes like here). But it should not occur here, plus target was clearly indicated as 2.18.

Oh well, no harm, it all worked out.

PR #249, opened 2024-07-26
android-record: Support deserializing polymorphic members of records

### LINKED ISSUE #248: jClass annotations and polymorphic types are ignored when deserializing Android Record fields
Suppose a class hierarchy like this:
```Java
public record Record(Field field) {
}

@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.PROPERTY, property="@class")
@JsonSubTypes({
        @JsonSubTypes.Type(value = StringField.class, name = "string"),
        @JsonSubTypes.Type(value = IntField.class, name = "int")
})
public abstract class Field {
}

public class StringField extends Field {

	private final String val;

	@JsonCreator
	public StringField(@JsonProperty("val") String val) {
		this.val = val;
	}

	public String getVal() {
		return val;
	}
}

public class IntField extends Field {

	private final int val;

	@JsonCreator
	public IntField(@JsonProperty("val") int val) {
		this.val = val;
	}

	public int getVal() {
		return val;
	}
}
```

Using Jackson on standard desktop Java SE, the following code works as expected:
```Java
String serialized = new ObjectMapper().writeValueAsString(r);
new ObjectMapper().readValue(serialized, Record.class);
```

However, that is not the case on Android (using the AndroidRecordModule), where the deserializer is oblivious of all class annotations (and, consequently, the JsonTypeInfo), and thus is unable to deserialize the abstract type `Field`. (`abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information`)

I would presume that the issue lies within this code: <https://github.com/FasterXML/jackson-modules-base/blob/9b2fe00f8f724749e6ba2c93efe1b26d3f33442b/android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java#L153>since in this case, neither the TypeDeserializer, nor getClassAnnotations are used for `Field`.

My current workaround is to wrap objects akin to `Field` in this example into the following:
```Java
public class ContainerObject<T> { //Android Studio's opinion of this code is "Class can be converted to record class". Well, if only I could:)
    public final T value;

    @JsonCreator
    public ContainerObject(@JsonProperty("value") T value) {
        this.value = value;
    }
}
```
and then change `Record` to be
```Java
public record Record(ContainerObject<Field> field) {
}
```
Using this wrapper, deserialization works as expected on Android. However, the general issue seems to be a bug within Jackson's AndroidRecordModule.

If so desired, I'm willing to fix this myself and PR it.

HelloOO7:
This fixes #248 (see tests). As far as I could tell, the non-default ValueInstantiators do not have access to a DeserializationContext when created, so there's a fair bit of redundancy in place when using `createContextual` instead.


eranl:
Thank you. LGTM. @cowtowncoder, your thougths?

cowtowncoder:
First of all, thank you @HelloOO7 !

Correct, context is passed via `createContextual()` same as is done for serializers and deserializers (for (de)serializers the reason are cyclic types, cycles cannot be resolved otherwise; and probably same is true for `ValueInstantiator`).

Code looks ok, although somehow it feels like this should be doable via `JsonDeserializer` side, and not on `ValueInstantiator`, similar to how regular JDK Record deserializer creation works.
But there may be reasons why this is not doable (`jackson-databind` might have access to something modules don't, for example).

Now: I really want to go through this in more detail, then merge, but am leaving for a vacation today. But will pick this up once I return in 2 weeks.
So don't worry when you won't here much until then -- this is one of top-5 items for me to do before 2.18.0-rc1 is finalized.

HelloOO7:
>  somehow it feels like this should be doable via JsonDeserializer side, and not on ValueInstantiator

I expanded upon ValueInstantiator in effort to adhere to the spirit of the original implementation. I may be wrong here, but wouldn't using JsonDeserializer needlessly couple the Android record module to the JSON format?

Upon looking into this further, I came up with a more similar-to-JDK approach to Android records that uses an AnnotationIntrospector, see here: https://github.com/HelloOO7/jackson-modules-base/tree/android-records-2
This approach is based on the fact that `POJOPropertiesCollector` handles records in `_addCreators` using:
```Java
if (_isRecordType) {
    primary = JDK14Util.findCanonicalRecordConstructor(_config, _classDef, constructors);
} else {
    primary = _annotationIntrospector.findDefaultCreator(_config, _classDef,
    constructors, factories);
}
```

However, this fails the `testDeserializeSimpleRecord_DisableAnnotationIntrospector` test (for apparent reasons). If that is a problem, it could probably be solved by overriding `_addCreators` itself in the `POJOPropertiesCollector` that the module overrides either way. It might also be of note that this change fixes 2 of the failing tests.

cowtowncoder:
I added included test in core `jackson-databind` to make sure it works with "real" Records as well (there was one test class for `@JsonTypeInfo` annotated members but coverage pretty slim).

Will try to get this reviewed now.

cowtowncoder:
> but wouldn't using JsonDeserializer needlessly couple the Android record module to the JSON format?

No, despite its name, `JsonDeserializer` is format agnostic (really should be `ValueDeserializer`).

cowtowncoder:
@HelloOO7 I like the looks of the alternate take -- could you do a PR for that instead, and I can get it quickly reviewed?

And yes, now that I looked at the original implementation it makes sense that you extended `ValueInstantiator` approach. But I much prefer `AnnotationIntrospector` approach, esp. with 2.18 addition of `findDefaultCreator`.

HelloOO7:
Sure. Would it be okay to merge the new commits to the main branch of my fork and update this PR, or should I submit a new one?

cowtowncoder:
@HelloOO7 Either way is fine, whatever is easiest to do?

HelloOO7:
Alright, I merged it to my branch of 2.18. It's now awaiting review as part of this PR as far as I can see (commit has appeared before my last message probably 'cause it's sorted chronologically).

HelloOO7:
I reckon I should probably update the tests to no longer be in the failing package, if this approch is the way to go forward.

cowtowncoder:
Hmmmh. I wish indentation was not changed -- merging this into `master` (3.0) is pretty painful due to bogus changes... :-(

cowtowncoder:
Thank you @HelloOO7  -- now merged in 2.18 for 2.18.0 as well as `master` for eventual 3.0.0.

HelloOO7:
Oh well. I tried to configure it to keep 2 spaces indentation (which I think it did), but the tests had strangely nested static classes on a different indentation level than static methods, which my IDE couldn't keep up with:( Sorry.

HelloOO7:
I'm really looking forward to have this fixed in 2.18. It's gonna be nice not to have to bother with workarounds. Thank you too!

cowtowncoder:
@HelloOO7 yeah np, indentation gets tricky no worries. Glad to have gotten it all merged.

PR #252, opened 2024-09-03
android-record: Adapt generic types returned by Android reflection to follow Java behavior

### LINKED ISSUE #251: android-record: Constructor is not recognized when a record uses both arrays and generic types
Given a record that contains both an array member and a generic type member, such as:
```Java
public record Test(
    int[] numberArray,
    Map<Integer, Integer> numberMap) {
}
```
this structure will fail to deserialize on Android with `no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator`.

I've researched this further and it seems like this is, in fact, an Android bug, and not a Jackson bug. If generic types are used (anywhere) in a class declaration, then Android's libcore parses the type descriptors in a way that [arrays are always instances of GenericArrayType](https://android.googlesource.com/platform/libcore/+/refs/heads/main/luni/src/main/java/libcore/reflect/GenericSignatureParser.java#293) and other types are always ParameterizedTypes. I'm almost certain this must be breaking some Java specification, and even if not, it is incredibly ugly. (for comedic effect, see [here](https://android.googlesource.com/platform/libcore/+/refs/heads/main/luni/src/main/java/libcore/reflect/GenericArrayTypeImpl.java#32) regarding how this behavior is worked around so that methods like "getGenericComponentType" work properly) (and for additional comedic effect, the average pay at Google is about $120K a year)

Since I reckon there's no point in trying to make Google fix this mess themselves (it has apparently been causing issues at least as early as [2018](https://github.com/kosi-libs/Kodein/issues/149)), I would rather propose to work around it in Jackson since it does break deserialization of quite a lot of realistic scenarios. I have made a fix for myself already and I can PR it if wanted, I would just need to rebase it to whatever point Jackson is currently at (it is currently based on where it was as of #248).

HelloOO7:
Fixes #251.

This PR adds a method to AndroidRecordModule that fixes/resolves GenericArrayTypes and ParameterizedTypes returned by Android reflection on method/constructor parameters. In turn, they behave the same as field types and constructor recognition on records that use both arrays and generic types no longer fails.

The type resolution method is compatible with the one in Android here: https://android.googlesource.com/platform/libcore/+/refs/heads/main/luni/src/main/java/libcore/reflect/ParameterizedTypeImpl.java#75
with the difference that it automatically resolves arrays fully where possible instead of returning an instance of GenericArrayType even on non-generic arrays (which causes errors in the current behaviour).

### DIFF
--- a/android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
+++ b/android-record/src/main/java/com/fasterxml/jackson/module/androidrecord/AndroidRecordModule.java
@@ -154,20 +154,21 @@ public class AndroidRecordModule extends SimpleModule
 
   static Type fixAndroidGenericType(Type type) {
     if (type instanceof GenericArrayType) {
-      //recurse into component type
       Type componentType = fixAndroidGenericType(((GenericArrayType) type).getGenericComponentType());
-      if (componentType instanceof Class<?>) { //if it isn't generic, return the raw array type
+      if (componentType instanceof Class<?>) {
         return arrayTypeCompat((Class<?>) componentType);
       }
     }
     else if (type instanceof ParameterizedType) {
       //if the parameterized type is not actually parameterized, deduce the raw type
       ParameterizedType parameterizedType = (ParameterizedType) type;
-      Type rawType = parameterizedType.getRawType();
-      if (rawType instanceof Class<?>) {
-        Class<?> rawComponentClass = (Class<?>) rawType;
-        if (rawComponentClass.getTypeParameters().length == 0) {
-          return rawComponentClass;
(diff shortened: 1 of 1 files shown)


cowtowncoder (approved):

cowtowncoder:
Forgot to ask for a unit test to show the fix, but not sure how feasible that'd have been. Merged for 2.18.0, regardless.

HelloOO7:
Oh yeah, unfortunately the behavior only occurs on Android, so I don't think it would be actually testable wirhout stubbing its reflection classes on desktop. Thank you.

cowtowncoder:
Right, that makes sense. Thank you for confirming.

PR #263, opened 2024-12-05
Fix to pass DeserializationContext along with Afterburner set calls

cowtowncoder:
(NOTE: not yet working)

As initial change to comply with newly added argument to `SettableBeanProperty.set()` (first arg, `DeserializationContext`) we pass just `null`. But ideally we would pass context through to generated mutator which might want to use it. This addition only applies to Jackson 3.0.

### DIFF
--- a/afterburner/src/main/java/tools/jackson/module/afterburner/deser/BeanPropertyMutator.java
+++ b/afterburner/src/main/java/tools/jackson/module/afterburner/deser/BeanPropertyMutator.java
@@ -1,5 +1,7 @@
 package tools.jackson.module.afterburner.deser;
 
+import tools.jackson.databind.DeserializationContext;
+
 /**
  * Abstract class that defines interface for implementations
  * that can be used proxy-like to change values of properties,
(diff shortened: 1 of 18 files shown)


cowtowncoder:
@geoand Not sure but I think you contributed Bytebuddy version of Afterburner code generation?

If so, was wondering if you could help here. The basic problem is that to conform to databind change

https://github.com/FasterXML/jackson-databind/issues/3561

there's need to pass in one more arguments via methods in `BeanPropertyMutator`: the newly added first argument of type `DeserializationContext`. I was trying to figure out where this occurs but couldn't quite do that (I did change some of argument counts as I think makes sense but am missing bytecode to actually copy incoming argument to out-going one).

geoand:
> @geoand Not sure but I think you contributed Bytebuddy version of Afterburner code generation?

A long time ago, yeah :)

> If so, was wondering if you could help here

I can take a look tomorrow and see if can figure it out

cowtowncoder:
@geoand Much appreciated -- hopefully it's something simple.

geoand:
@cowtowncoder I have a fix. How would you want me to get it into this PR? 

In any case I am attaching the patch.
[Fix_needed_for_use_of_DeserializationContext.txt](https://github.com/user-attachments/files/18037753/Fix_needed_for_use_of_DeserializationContext.txt)

cowtowncoder:
@geoand thank you -- let me see if I can just patch that in tonight.

cowtowncoder:
Thank you @geoand -- it merged cleanly, so update PR & merged!

geoand:
You are welcome!

Issue #268, opened 2025-01-11
Unify testing structure/tools [JSTEP-10]


JooHyukKim:
This issue serves as an umbrella issue for resolving [[JSTEP-10: Unifying testing structure/tools]](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-10)

### TODO

- Add JUnit 5 dep + `failing` test utils
- Migrate modules to JUnit 5
- Migrate failing tests to `/tofix` and remove existing `/failing` funciontality
- Remove JUnit 4

cowtowncoder:
So far converted:

* `afterburner`
* `android-record`
* `guice`
* `guice7`

Biggest remaining work is `blackbird`

cowtowncoder:
And with @JooHyukKim's work, all the rest -- Blackbird, jakarta-xmlbind, jaxb, paranamer, OSGi -- also converted.

What remains is converting "failing" tests to "tofix" ones, and then promoting JUnit5 dependencies from sub-modules' `pom.xml`s back to parent pom.xml.

cowtowncoder:
Whops. Still need to "mrbean".

cowtowncoder:
Status:

* All modules converted to use JUnit5
* JUnit5 dependency at main level (no per-module deps)

so what remains is just

* Convert "failing" tests to "tofix" ones.

JooHyukKim:
Filed last remaining pieces in two PR's "in order"! 👍🏼

cowtowncoder:
Thanks! Will get these merged today.

cowtowncoder:
Merged, closing.

PR #271, opened 2025-01-17
Start work on #268: junit5 conversion; first demote junit4 dependency

cowtowncoder:
(needed so we can convert modules one by one and possibly leave some on JUnit 4 if need be)


cowtowncoder:
@JooHyukKim After making base modules properly modular, I'll start incremental work on JUnit 5 conversion. Due to challenges in merging, will do multiple PRs per phase.

PR #282, opened 2025-02-04
Fix META-INF/services definitions (before removing from 3.0)

cowtowncoder:


cowtowncoder:
Module metadata invalid since still used 2.x packages, class name.

Issue #285, opened 2025-02-17
`JacksonBlackbirdAccess` Class access exception with Groovy Script Serialization using `BlackbirdModule`


apoorvam:
I'm encountering an issue when attempting to use the BlackbirdModule in a Spring Boot 3.2 application with Groovy scripting. The goal is to serialize a Groovy object into JSON using the Jackson ObjectMapper with the BlackbirdModule. However, the test case below fails with an exception during the serialization of a Groovy object.

```java
 @Test
    void testGroovySerializationWithBlackbird() throws Exception {
        // Define a Groovy script as a String
        String script = """
                    import com.fasterxml.jackson.databind.ObjectMapper
                    import com.fasterxml.jackson.module.blackbird.BlackbirdModule

                    class Person {
                        String name
                        int age
                    }

                    def mapper = new ObjectMapper()
                    mapper.registerModule(new BlackbirdModule())

                    def person = new Person(name: "Groovy User", age: 35)
                    def json = mapper.writeValueAsString(person)

                    return json // Returning the JSON string
                """;

        // Execute the Groovy script
        GroovyShell shell = new GroovyShell();
        Object result = shell.evaluate(script);

        assertNotNull(result);
    }
```

Exception: 
```
java.lang.ClassFormatError: Illegal class name "/$$JacksonBlackbirdAccess" in class file /$$JacksonBlackbirdAccess
	at java.base/java.lang.ClassLoader.defineClass0(Native Method)
	at java.base/java.lang.System$2.defineClass(System.java:2394)
	at java.base/java.lang.invoke.MethodHandles$Lookup$ClassDefiner.defineClass(MethodHandles.java:2505)
	at java.base/java.lang.invoke.MethodHandles$Lookup$ClassDefiner.defineClass(MethodHandles.java:2480)
	at java.base/java.lang.invoke.MethodHandles$Lookup.defineClass(MethodHandles.java:1865)
	at com.fasterxml.jackson.module.blackbird.CrossLoaderAccess.accessClassIn(CrossLoaderAccess.java:131)
	at com.fasterxml.jackson.module.blackbird.CrossLoaderAccess.grantAccess(CrossLoaderAccess.java:94)
	at com.fasterxml.jackson.module.blackbird.CrossLoaderAccess.apply(CrossLoaderAccess.java:83)
	at com.fasterxml.jackson.module.blackbird.CrossLoaderAccess.apply(CrossLoaderAccess.java:11)
[...]

PR #289, opened 2025-03-07
Add test to reproduce #285

apoorvam:


cowtowncoder:
@apoorvam since it's your first contribution, I had to manually approve running of CI -- it would normally always run (and will do so in future).

blackbird/src/test/java/com/fasterxml/jackson/module/blackbird/TestNoPackageSerialization.java
```diff
@@ -0,0 +1,78 @@
+package com.fasterxml.jackson.module.blackbird;
```
cowtowncoder, line 1: Since it's serialization test, move to `com.fasterxml.jackson.module.blackbird.ser`

cowtowncoder:
Ok makes sense: couple of suggestions for changes.

But... why does it pass, without fix?

cowtowncoder:
Assuming all goes well @apoorvam  -- and thank you again for your work here!!! -- one thing we'll need before merging PR(s) is CLA:

it only needs to be sent once, before the first contribution & is good from there on for any other PRs.

The usual way is to print the doc, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
And once we get that I can proceed with merge (pending code review etc)

Thank you again!

cowtowncoder (approved):

PR #310, opened 2025-09-08
Update LICENSE for jackson-module-blackbird

JKomoroski:
Fixes the packaged license file to reference the module being packaged.

Looks like this may have been overlooked when blackbird was created from the old afterburner impl.


cowtowncoder (approved):

cowtowncoder:
Thank you @JKomoroski -- definitely copy-paste issue.

Issue #355, opened 2026-07-17
Blackbird silently reverts Optional<T> (reference-type) properties to reflection MethodProperty: Jdk8OptionalDeserializer fails isDefaultDeserializer check


pavel-ptashyts:
## Environment

- `tools.jackson.core:jackson-core` / `jackson-databind` **3.1.2**, `jackson-module-blackbird` 3.1.2 (Jackson 3 line)
- JDK 21 (Temurin), Linux x64
- Also checked 3.1.3–3.1.5, 3.2.0, 3.2.1 sources and release notes — the behavior described below is unchanged.

## Summary

When Blackbird is registered, bean properties whose type is `java.util.Optional<T>` (and, more generally, any reference-type property handled by `ReferenceTypeDeserializer`) are first optimized and then **silently reverted to the reflection-based `MethodProperty`** during deserializer resolution. The result is that on Optional-heavy DTOs Blackbird quietly provides no benefit for most properties, and there is no diagnostic hint that this happened.

## Root cause

Two phases:

1. `BBDeserializerModifier.updateBuilder(...)` runs **before** deserializer resolution. At that point `prop.hasValueDeserializer()` is `false`, the raw parameter type check passes (`Optional` is a plain object type), and the property is optimized into `SettableObjectProperty` with a `LambdaMetafactory`-generated `BiConsumer` setter. So far so good.

2. Later, `BeanDeserializerBase.resolve()` assigns the value deserializer: `prop.withValueDeserializer(...)` with `tools.jackson.databind.ext.jdk8.Jdk8OptionalDeserializer`. This lands in Blackbird's `OptimizedSettableBeanProperty.withValueDeserializer`:

```java
SettableBeanProperty newDelegate = delegate.withValueDeserializer(deser);
if (!_isDefaultDeserializer(deser)) {
    return newDelegate;          // <-- silently reverts to plain MethodProperty
}
```

with

```java
protected boolean _isDefaultDeserializer(ValueDeserializer<?> deser) {
    return ClassUtil.isJacksonStdImpl(deser)
            || (deser instanceof SuperSonicBeanDeserializer);
}
```

[...]

PR #356, opened 2026-07-20
Keep Blackbird optimization for built-in Optional deserializer

dlwldnjs1009:
Addresses #355. Blackbird treats `Jdk8OptionalDeserializer` as custom when used with databind versions that predate its `@JacksonStdImpl` marker, replacing the optimized property with `MethodProperty` during deserializer resolution.

Changes:
- Recognize the exact `Jdk8OptionalDeserializer` class as a built-in handler
- Verify built-in Optional properties retain `SettableObjectProperty`
- Verify user subclasses continue to fall back to `MethodProperty`

The exact-class check avoids broadening the predicate to all `ReferenceTypeDeserializer` implementations, so custom subclasses retain the existing reflection fallback. FasterXML/jackson-databind#6109 adds marker-based classification for marker-bearing databind versions; this check covers pre-marker 3.1 patch releases.

Testing:
- `./mvnw -pl blackbird test -Dtest=OptionalDeserTest`
- `./mvnw -pl blackbird test`

### DIFF
--- a/blackbird/src/main/java/tools/jackson/module/blackbird/deser/OptimizedSettableBeanProperty.java
+++ b/blackbird/src/main/java/tools/jackson/module/blackbird/deser/OptimizedSettableBeanProperty.java
@@ -8,6 +8,7 @@ import tools.jackson.core.*;
 import tools.jackson.databind.*;
 import tools.jackson.databind.deser.*;
 import tools.jackson.databind.deser.impl.NullsConstantProvider;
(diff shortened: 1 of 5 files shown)


cowtowncoder:
I think this makes sense for 3.1(.6); will merge there, 3.2(.2), but drop from 3.x (3.3.0) trusting `@JacksonStdImpl` to work fully.

Issue #10, opened 2016-05-07
Afterburner failing inside Karaf 3.0.5


debraj-manna:
I have a simple Hello World type Karaf Bundle. The activator code looks like below:-

```
public class HelloWorldActivator implements BundleActivator {
	 	@Override
	    public void start(BundleContext bundleContext) throws Exception {
	        System.out.println("STARTING DEMO: hello, world\n");
	        System.out.println(getJsonDataAsString());
	    }
	 
	    @Override
	    public void stop(BundleContext bundleContext) throws Exception {
	        System.out.println("STOPPING DEMO");
	    }	
	    
	    private String getJsonDataAsString() {
	    	JsonDataBlob jsonDataBlob = new JsonDataBlob();
	    	ObjectMapper objectMapper = new ObjectMapper();
	    	objectMapper.registerModule(new AfterburnerModule());
	    	try {
	    		return objectMapper.writeValueAsString(jsonDataBlob);
	    	} catch(Exception e) {
	    		e.printStackTrace();
	    	}
	    	return "";	    	
	    }	    	    
}
```

`pom.xml` looks like below:-

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>com.hello</groupId>
	<artifactId>world</artifactId>
	<version>0.0.1</version>
	<packaging>bundle</packaging>
	<name>Hello World</name>
	<dependencies>
		<!-- OSGi -->
		<dependency>
			<groupId>org.osgi</groupId>
			<artifactId>org.osgi.core</artifactId>
			<version>4.3.1</version>
		</dependency>
		<dependency>
			<groupId>org.osgi</groupId>
			<artifactId>org.osgi.compendium</artifactId>
			<version>4.3.1</version>
		</dependency>
		<dependency>
			<groupId>org.json</groupId>
			<artifactId>json</artifactId>
			<version>20160212</version>
		</dependency>
		<dependency>
			<groupId>com.fasterxml.jackson.core</groupId>
			<artifactId>jackson-core</artifactId>
			<version>2.6.2</version>
		</dependency>

		<dependency>
			<groupId>com.fasterxml.jackson.core</groupId>
[...]

PR #359, opened 2026-09-07
Blackbird: ClassFile generation of unrolled codec class, optimize record / builder / field

### LINKED ISSUE #142: `WrongMethodTypeException` with blackbird
After upgrading to Java 16 and Jackson 2.12 with Blackbird replacing Afterburner, we are seeing WrongMethodTypeException occur frequently. The strange thing is that this affects only 25-40% of requests. An identical request made repeatedly will fail at that rate.

Example stack trace:

```
java.lang.invoke.WrongMethodTypeException: expected ()ToIntFunction but found ()ToIntFunction
	at java.base/java.lang.invoke.Invokers.newWrongMethodTypeException(Invokers.java:506)
	at java.base/java.lang.invoke.Invokers.checkExactType(Invokers.java:515)
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.createProperty(BBSerializerModifier.java:116)
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.lambda$findProperties$0(BBSerializerModifier.java:64)
	at com.fasterxml.jackson.module.blackbird.util.Unchecked.lambda$runnable$0(Unchecked.java:31)
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.findProperties(BBSerializerModifier.java:65)
	at com.fasterxml.jackson.module.blackbird.ser.BBSerializerModifier.changeProperties(BBSerializerModifier.java:49)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.constructBeanOrAddOnSerializer(BeanSerializerFactory.java:407)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.findBeanOrAddOnSerializer(BeanSerializerFactory.java:290)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory._createSerializer2(BeanSerializerFactory.java:235)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.createSerializer(BeanSerializerFactory.java:169)
	at com.fasterxml.jackson.databind.SerializerProvider._createUntypedSerializer(SerializerProvider.java:1473)
	at com.fasterxml.jackson.databind.SerializerProvider._createAndCacheUntypedSerializer(SerializerProvider.java:1421)
	at com.fasterxml.jackson.databind.SerializerProvider.findContentValueSerializer(SerializerProvider.java:753)
	at com.fasterxml.jackson.databind.ser.impl.PropertySerializerMap.findAndAddSecondarySerializer(PropertySerializerMap.java:90)
	at com.fasterxml.jackson.databind.ser.std.AsArraySerializerBase._findAndAddDynamic(AsArraySerializerBase.java:311)
	at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serializeContents(IndexedListSerializer.java:115)
	... 74 common frames omitted
```

This is occuring in a Dropwizard application.

Happy to provide more information as needed, and will continue to dig and update this issue with what I find.

### LINKED ISSUE #350: Remove CrossLoaderAccess companion-class slow path from Blackbird (deprecated in 3.2)
Follow-up to #349 (blackbird-tests module). Deprecated in 3.2, targeting removal in **3.3**.

## What to remove

In `blackbird/src/main/java/tools/jackson/module/blackbird/CrossLoaderAccess.java`:

- `private static final String CLASS_NAME` — the `$$JacksonBlackbirdAccess` constant
- `private static final int[] HEADER` — pre-compiled Java 8 bytecode template
- `private static final int[] FOOTER` — pre-compiled Java 8 bytecode template
- `private static Class<?> accessClassIn(MethodHandles.Lookup)` — the method that defines the companion class at runtime via `Lookup.defineClass(byte[])`
- The fall-through branch in `grantAccess`, after which `grantAccess` simplifies to:
  ```java
  private static MethodHandles.Lookup grantAccess(MethodHandles.Lookup lookup) {
      return lookup;
  }
  ```
  (or remove `grantAccess` entirely and make `CrossLoaderAccess.apply` the identity function)

All four members are already annotated `@Deprecated(since = "3.2", forRemoval = true)` with javadoc pointing at this issue's rationale.

## Why it's safe to remove

The slow path is **never reached on JDK 9+ for any real input**:

1. `BBDeserializerModifier.updateBuilder` and `BBSerializerModifier.findProperties` both call `ReflectionHack.privateLookupIn(beanClass, lookup)` before delegating to `CrossLoaderAccess.apply`.
2. On JDK 9+, `MethodHandles.privateLookupIn` always returns a lookup with `hasFullPrivilegeAccess() == true` whenever it succeeds. And it succeeds for every bean class Blackbird can see (including classpath / unnamed-module beans — the unnamed module is considered \"exported to all\" for `privateLookupIn` purposes).
3. `CrossLoaderAccess.grantAccess` short-circuits as soon as `hasFullAccess(lookup)` is true, returning the lookup unchanged and never reaching the `accessClassIn(...)` call.
4. Blackbird's 3.x baseline is Java 17, so the `DEFINE_CLASS == null` branch (the Java 8 fallback) is also unreachable.

The slow path was written in the Java 8 era when `privateLookupIn` didn't exist and callers had to synthesize a full-privilege lookup by defining a companion class in the target package. That scaffolding is no longer needed.

## Verification

`blackbird-tests/src/test/java/tools/jackson/module/blackbird/inject/CrossLoaderAccessTest.java` (added in #349) pins the current behavior:

```java
@Test
public void testFastPathWins_NoCompanionClassDefined() throws Exception {
    Harness h = newHarness();
    XLoaderBean bean = h.mapper.readValue(\"{\\\"a\\\":1,\\\"b\\\":\\\"hi\\\"}\", XLoaderBean.class);
    // ... setters are optimized, deserialization works ...
    assertFalse(companionClassExists(),
            \"CrossLoaderAccess unexpectedly defined $$JacksonBlackbirdAccess ...\");
}
```

This test currently passes on JDK 17. The removal PR should keep the test passing — if removing the slow path makes the test fail, that would be evidence the slow path was actually being hit and the deprecation was wrong; in that case the fix is to keep the method, remove the `@Deprecated` marker, and update the test to cover whichever scenario was hitting the slow path.

During the 3.2 → 3.3 development cycle, before starting the removal:

1. Run `blackbird-tests/CrossLoaderAccessTest` on JDK 17, 21, and whatever is the latest LTS at that time. Confirm the fast-path assertion still holds.
2. Grep for any downstream reports of classes named `$$JacksonBlackbirdAccess` appearing in production — that would be an existence proof that the slow path is live somewhere we haven't mapped.
3. If any user has overridden `BlackbirdModule.findLookup()` or `findLookupSupplier()` to return a non-full-privilege `Lookup` (e.g. `MethodHandles.publicLookup()` or a deliberately reduced one), document the requirement that the returned lookup must satisfy `hasFullPrivilegeAccess()` after `privateLookupIn` — or add a runtime check that throws with a clear message, rather than silently degrading.

## Scope

Small PR: ~100 lines deleted (most of them are the `HEADER` / `FOOTER` byte arrays), no new tests required (existing `CrossLoaderAccessTest` already covers the post-removal behavior), release notes entry.

## Related

- #349 — blackbird-tests module; where the dead-code finding was surfaced
- #347 — afterburner-tests; the companion module whose investigation led to the analogous finding being applied to Blackbird
- #348 — afterburner classloader caching regression; different bug, different module, unrelated mechanism — Blackbird is structurally immune because it never uses reflective `ClassLoader.defineClass` to begin with

stevenschlansker:
This replaces the Blackbird implementation. The module coordinates, the `BlackbirdModule` API, and the registration pattern do not change. The LambdaMetafactory engine is gone; the module now generates one hidden class per bean type with the ClassFile API (JEP 484) and installs it as that bean's deserializer and serializer.

Why replace rather than extend: Jackson 3 core ships `PropertyNameMatcher` and `JsonParser.nextNameMatch()`, which match property names from the input buffer and return an index. A generated codec can consume that index in a `tableswitch` and read each property with direct, monomorphic code. The LambdaMetafactory design cannot express this; it only accelerates the accessor calls inside the stock deserializer loop, and the measured gap shows most of the remaining cost sits in that loop, not in the accessors.

Against old Blackbird 3.2.2 at this PR head, the rewrite reads records 59% faster (2.2x vanilla databind), builder beans 28% faster, and setter POJOs 25% faster, with writes up 4% to 15%

**This PR requires modules-base project to be released with JDK 25, although it maintains JDK 17 support for all modules but Blackbird**

Designed with Claude Fable 5. Claude's notes below:

### The engine

- One hidden class per bean type, defined with `defineHiddenClassWithClassData`. classData carries the name matcher, child codec handles, constructor or build-method handles, and the stock deserializers used as fallbacks. Constant MethodHandles + `invokeExact` inline like direct calls.
- Hidden classes are defined without `ClassOption.STRONG`. The codec instance in the mapper's cache anchors the class, so codecs unload with the mapper. This fixes the documented Blackbird classloader-pinning problem (#147): the old engine's lambdas stayed alive with the target ClassLoader, so applications that create many mappers ran out of metaspace. The README warning that #215 added for this comes out.
- Construction modes: setter POJOs (direct `new` plus `invokevirtual` setters), records (typed locals in canonical-constructor order, one constructor call), and builder beans (stock `ValueInstantiator` creates the builder, fluent setters inline, build method through a constant handle). Afterburner and old Blackbird never accelerated the record or builder paths.
- Per-property tiers, gated on the effective configuration: simple scalar types read with direct parser calls; nested beans call the child codec through a classData handle; everything else (custom deserializers, coercion or null handling that is not default, polymorphic typing, injectables, and similar) routes through the stock `SettableBeanProperty`, so behavior matches stock databind. Beans the generator cannot cover take the whole-bean bailout and keep the stock `BeanDeserializer`.
- Serializer codegen is the same design in reverse: straight-line writers with pre-encoded name constants, always on like the deserializer side; see "Serializer default" below for the C2 history behind the write helpers.
- Views handled as first class citizens

### What Accelerates

Beyond the base rewrite, the following now accelerate: beans with `@JsonAnySetter` (POJO and builder; the unknown arm feeds the stock any-setter in `handleUnknownVanilla`'s exact order); POJOs whose default creator is a no-arg `@JsonCreator` factory or a custom instantiator (construct through the stock `ValueInstantiator`); beans with `@JacksonInject` (POJO and builder; injectors apply after construction, before the property loop, so document values override injected ones like stock); and polymorphic writes - `serializeWithType` is generated natively with stock `BeanSerializerBase`'s `WritableTypeId` flow for every inclusion mechanism (PROPERTY, WRAPPER_OBJECT, WRAPPER_ARRAY). Records with an any-setter, injectables, or a factory creator, and beans with a `@JsonTypeId` property, stay on the stock path (pinned).

 ### Measured results
  ```
  Results (x86_64 qa pod, JDK 26, medians of 6 rotated rounds)
  
  ┌──────────────────────────┬─────────┬────────┬───────┬─────────┬─────────────┐
  │           Cell           │ vanilla │ old BB │  new  │ new/old │ new/vanilla │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ order read (record)      │ 188k    │ 257k   │ 409k  │ +59%    │ 2.17x       │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ invoice read (builder)   │ 296k    │ 297k   │ 381k  │ +28%    │ +29%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ media read (setter POJO) │ 259k    │ 265k   │ 331k  │ +25%    │ +28%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ field read               │ 664k    │ 665k   │ 770k  │ +16%    │ +16%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ poly read                │ 913k    │ 944k   │ 1040k │ +10%    │ +14%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ catalog read             │ 263k    │ 262k   │ 277k  │ +5%     │ +5%         │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ field write              │ 1065k   │ 1058k  │ 1222k │ +15%    │ +15%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ order write              │ 568k    │ 637k   │ 687k  │ +8%     │ +21%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ media write              │ 448k    │ 478k   │ 496k  │ +4%     │ +11%        │
  ├──────────────────────────┼─────────┼────────┼───────┼─────────┼─────────────┤
  │ poly write               │ 1405k   │ 1596k  │ 1619k │ +1%     │ +15%        │
  └──────────────────────────┴─────────┴────────┴───────┴─────────┴─────────────┘

```

### Access

The rewrite no longer needs a user-supplied `MethodHandles.Lookup` for any acceleration. Generated codecs reach members through `MethodHandle` constants unreflected with the module's own lookup after databind has already run `fixAccess` on those members, and every descriptor is erased (references widened to `Object`), so the generated bytecode never names the bean class.
Consequences: private classes, private members, private record constructors, private builder build methods, and foreign-classloader beans all accelerate; every codec defines in the module's own context; on the module path the only requirement is the same `opens ... to tools.jackson.databind` stock databind already needs. The `BlackbirdModule` lookup constructors and `findLookup`/`findLookupSupplier` overrides remain as released API but are no longer required by anything. This deletes `CodecAccess`, the `privateLookupIn` define-context machinery, and the foreign-classloader gate.

Fixes #142 - the affected code path is removed.

### Serializer inlining

Generated straight-line writers initially lost ~24% to the old engine on aarch64 number-heavy record shapes. The cause is an AArch64 C2 effect: when small hot jackson-core methods (`NumberOutput.outputInt`/`outputLong`, `UTF8JsonGenerator.writeName`) inline into a looping writer body, the register allocator spills working state across the loop. The generator now emits per-kind static write helpers sized past the C2 inline threshold, so those copies compile standalone; with that change the generated writers beat the old engine on the regressing shape too (+9% aarch64 record writes, +18% media writes, and the run-to-run bimodality is gone; JDK 25 and 26). The helper-based build was revalidated on x86_64: it wins or ties every cell there as well (media +9% over the old engine, all rounds; order +4% median).

### Compatibility notes

- JDK floor moves to 25 (`--release 25`; the ClassFile API is final in 24, and 25 is the LTS). Classic Blackbird continues to serve older JDKs, but otherwise is considered retired.
- Records, fields and builder beans now accelerate; old Blackbird left all on the stock path.
- Beans outside the generator's coverage keep stock behavior by construction: the modifier returns the stock deserializer unchanged.
- Native-image detection is unchanged (the module deactivates itself).
- One pre-existing test-infrastructure note: tofix/TestBBClassloaders passes on the classpath and fails on the module path on pristine 3.x, so the failure-expected annotation is mode-dependent. This predates the rewrite.

### Removed dead code

After both engines were replaced, `CrossLoaderAccess` survived only as an unused constructor parameter and the `util` chain (`ReflectionHack`, `Unchecked`, `Sneaky`, `CheckedFunction`, `CheckedSupplier`) had no callers.
This PR deletes them (-504 lines) and closes #350, which scheduled the CrossLoaderAccess removal for 3.3 (deprecated `forRemoval` since 3.2).

## Appendix: examples of generated code

Each engaged bean gets one hidden class per direction, generated with the
ClassFile API and defined through `defineHiddenClassWithClassData`. The class
extends `GeneratedReaderBase` (deserializer) or `GeneratedWriterBase`
(serializer) and receives its constants - the property-name matcher, the
stock `SettableBeanProperty`/`PropertyWriter` fallbacks per property, child
codecs, pre-encoded names, and the member-access and construction
`MethodHandle`s - through a name-addressed classData map. The decompiler
renders each entry as the pseudocode `classDataEntry<"idProp">()`: in the
class file it is a named dynamic constant (condy), resolved once and
constant-foldable by the JIT, not a per-call lookup.

Member access and construction go through `MethodHandle` constants with
erased descriptors: every reference type is widened to `Object`, primitives
kept, so the generated bytecode never names the bean class. A setter reads as
`classDataEntry<"mediaSet">().invokeExact((Object)bean, (Object)value)`, a
constructor as `classDataEntry<"constructor">().invokeExact(...)`, a getter as
`(int)classDataEntry<"widthGet">().invokeExact((Object)value)`. The handles
are unreflected with the module's own lookup after databind has run
`fixAccess` on the members, so this reaches exactly what stock databind can
invoke - private members, non-public classes, records, foreign classloaders -
and every codec defines in the module's own package context regardless of the
bean's module or loader. A condy `MethodHandle` plus `invokeExact` inlines
like a direct call, so the erased form costs nothing at steady state.

Generated methods carry parameter names and a LocalVariableTable, so
decompiled bodies read with source-like names; reference record components and
the bean receiver appear as `Object` because their slots are erased. Views: a
bean that declares `@JsonView` resolves a cached per-view visibility bitmask
at entry and each property arm tests its bit, so view-active calls stay on the
generated path; the beans below declare none, so they show only the entry
guard, which hands non-`START_OBJECT` entries - and, where a view could
matter, view-active calls - to the stock fallback.

Two decompiler artifacts to read past: the bytecode's entry guard branches
forward to a delegation tail after the body (early-out shape), but the
decompiler renders it as a nested conditional; and the per-arm exception
regions all share one handler (a topology javac never emits), which the
decompiler expresses with synthetic `boolean varNN = false` routing flags and
its "$VF: Inserted dummy exception handlers" banner - the class file's
handler is a plain four-instruction rethrow through `_propertyException`,
with no flags and no dead stores.

The examples below are decompiled with Vineflower 1.10.1 from bytes dumped by
the module itself (`-Dblackbird.debug.dumpDir`, see Reproducing below), at the
follow-up-wave HEAD, and are unedited except for the marked elisions. The
shapes come from the benchmark corpus.

<details>
<summary><b>Record deserializer</b> - <code>Order</code> (nested records, child codec, list property; erased construction handle; shown in full)</summary>

The bean:

```java
public record Order(String id, long amountCents, boolean expedited, Customer customer, List<Line> lines) {
    public record Customer(String name, String email, int tier) {}
    public record Line(String sku, int quantity, long priceCents) {}
}
```

The generated reader (`bbnext.model.Order-reader.class`): components read
into typed locals under their component names (reference components as
`Object`, since the constructor handle is erased) and the canonical
constructor is invoked once through `classDataEntry<"constructor">`; each
scalar arm checks the expected token and demotes anything else to the stock
property; the nested `customer` calls the child codec only for `START_OBJECT`;
the `lines` list rides its stock property; `seen` tracks components for
required-property and `FAIL_ON_MISSING_CREATOR_PROPERTIES` reporting, and a
null mask feeds `FAIL_ON_NULL_CREATOR_PROPERTIES`; unknown names go through
`_handleUnknown`; a property failure is rethrown with the stock reference path
via `_propertyException`; anything but `START_OBJECT` at entry, or an active
view, delegates to the stock deserializer.

```java
package tools.jackson.module.blackbird.codegen;

import java.lang.invoke.MethodHandle;
import java.util.Set;
import tools.jackson.core.JsonParser;
import tools.jackson.core.JsonToken;
import tools.jackson.core.sym.PropertyNameMatcher;
import tools.jackson.databind.DeserializationContext;
import tools.jackson.databind.deser.SettableBeanProperty;
import tools.jackson.databind.deser.bean.BeanDeserializerBase;
import tools.jackson.module.blackbird.internal.GeneratedReaderBase;

public final class BBReader_Order extends GeneratedReaderBase {
   public BBReader_Order(BeanDeserializerBase fallback, boolean ignoreAllUnknown, Set ignorableProps, Set includableProps) {
      super(fallback, ignoreAllUnknown, ignorableProps, includableProps);
   }

   // $VF: Inserted dummy exception handlers to handle obfuscated exceptions
   public Object deserialize(JsonParser p, DeserializationContext ctxt) {
      if ((p.currentToken() == JsonToken.START_OBJECT || p.currentToken() == JsonToken.PROPERTY_NAME) && ctxt.getActiveView() == null) {
         Object id = null;
         long amountCents = 0L;
         boolean expedited = false;
         Object customer = null;
         Object lines = null;
         long seen = 0L;
         PropertyNameMatcher matcher = (PropertyNameMatcher)GeneratedReaderBase.classDataEntry<"matcher">();
         int ix;
         if (p.currentToken() != JsonToken.PROPERTY_NAME) {
            ix = p.nextNameMatch(matcher);
         } else {
            ix = p.currentNameMatch(matcher);
         }

         while (true) {
            while (ix < 0) {
               if (ix == -1) {
                  if (seen != 31L) {
                     this._checkRecordSeen(ctxt, seen, 5);
                  }

                  long var17 = 0L;
                  if (id == null) {
                     var17 |= 1L;
                  }

                  if (customer == null) {
                     var17 |= 8L;
                  }

                  if (lines == null) {
                     var17 |= 16L;
                  }

                  if (var17 != 0L) {
                     this._checkRecordNulls(ctxt, var17, 5);
                  }

                  return (Object)(MethodHandle)GeneratedReaderBase.classDataEntry<"constructor">()
                     .invokeExact((Object)id, (long)amountCents, (boolean)expedited, (Object)customer, (Object)lines);
               }

               if (ix != -2) {
                  return this._unexpectedToken(p, ctxt);
               }

               this._handleUnknown(p, ctxt, null);
               ix = p.nextNameMatch(matcher);
            }

            SettableBeanProperty prop;
            Exception var10000;
            switch (ix) {
               case 0:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"idProp">();

                  try {
                     if (p.nextToken() == JsonToken.VALUE_STRING) {
                        id = p.getString();
                     } else {
                        id = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"idProp">().deserialize(p, ctxt);
                     }
                  } catch (Exception var23) {
                     var10000 = var23;
                     boolean var28 = false;
                     break;
                  }

                  seen |= 1L;
                  ix = p.nextNameMatch(matcher);
                  continue;
               case 1:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"amountCentsProp">();

                  try {
                     if (p.nextToken() == JsonToken.VALUE_NUMBER_INT) {
                        amountCents = p.getLongValue();
                     } else {
                        amountCents = (Long)(SettableBeanProperty)GeneratedReaderBase.classDataEntry<"amountCentsProp">().deserialize(p, ctxt);
                     }
                  } catch (Exception var22) {
                     var10000 = var22;
                     boolean var27 = false;
                     break;
                  }

                  seen |= 2L;
                  ix = p.nextNameMatch(matcher);
                  continue;
               case 2:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"expeditedProp">();

                  try {
                     label95: {
                        JsonToken var24 = p.nextToken();
                        if (var24 != JsonToken.VALUE_TRUE) {
                           if (var24 != JsonToken.VALUE_FALSE) {
                              expedited = (Boolean)(SettableBeanProperty)GeneratedReaderBase.classDataEntry<"expeditedProp">().deserialize(p, ctxt);
                              break label95;
                           }
                        }

                        expedited = p.getBooleanValue();
                     }
                  } catch (Exception var21) {
                     var10000 = var21;
                     boolean var26 = false;
                     break;
                  }

                  seen |= 4L;
                  ix = p.nextNameMatch(matcher);
                  continue;
               case 3:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"customerProp">();

                  try {
                     if (p.nextToken() == JsonToken.START_OBJECT) {
                        customer = (GeneratedReaderBase)GeneratedReaderBase.classDataEntry<"customerCodec">().deserialize(p, ctxt);
                     } else {
                        customer = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"customerProp">().deserialize(p, ctxt);
                     }
                  } catch (Exception var20) {
                     var10000 = var20;
                     boolean var25 = false;
                     break;
                  }

                  seen |= 8L;
                  ix = p.nextNameMatch(matcher);
                  continue;
               case 4:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"linesProp">();

                  try {
                     p.nextToken();
                     lines = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"linesProp">().deserialize(p, ctxt);
                  } catch (Exception var19) {
                     var10000 = var19;
                     boolean var10001 = false;
                     break;
                  }

                  seen |= 16L;
                  ix = p.nextNameMatch(matcher);
                  continue;
               default:
                  throw new IllegalStateException("bad property index");
            }

            Exception e = var10000;
            throw this._propertyException(e, null, prop, ctxt);
         }
      } else {
         return super._fallback.deserialize(p, ctxt);
      }
   }
}
```
</details>

<details>
<summary><b>Setter-POJO deserializer</b> - <code>MediaContent</code> (erased constructor and setter handles, a child codec, a stock list arm; shown in full)</summary>

The generated reader constructs through the no-arg constructor handle
(`classDataEntry<"constructor">().invokeExact()` returning `Object`), sets the
nested `media` through an erased setter handle
(`classDataEntry<"mediaSet">().invokeExact((Object)bean, (Object)value)`) when
the value is a `START_OBJECT` the child codec can read, and rides the stock
property for the `images` list. The bean local is `Object`; no bean-class name
appears.

```java
package tools.jackson.module.blackbird.codegen;

import java.lang.invoke.MethodHandle;
import java.util.Set;
import tools.jackson.core.JsonParser;
import tools.jackson.core.JsonToken;
import tools.jackson.core.sym.PropertyNameMatcher;
import tools.jackson.databind.DeserializationContext;
import tools.jackson.databind.deser.SettableBeanProperty;
import tools.jackson.databind.deser.bean.BeanDeserializerBase;
import tools.jackson.module.blackbird.internal.GeneratedReaderBase;

public final class BBReader_MediaContent extends GeneratedReaderBase {
   public BBReader_MediaContent(BeanDeserializerBase fallback, boolean ignoreAllUnknown, Set ignorableProps, Set includableProps) {
      super(fallback, ignoreAllUnknown, ignorableProps, includableProps);
   }

   // $VF: Inserted dummy exception handlers to handle obfuscated exceptions
   public Object deserialize(JsonParser p, DeserializationContext ctxt) {
      if ((p.currentToken() == JsonToken.START_OBJECT || p.currentToken() == JsonToken.PROPERTY_NAME) && ctxt.getActiveView() == null) {
         Object bean = (Object)(MethodHandle)GeneratedReaderBase.classDataEntry<"constructor">().invokeExact();
         p.assignCurrentValue(bean);
         PropertyNameMatcher matcher = (PropertyNameMatcher)GeneratedReaderBase.classDataEntry<"matcher">();
         int ix;
         if (p.currentToken() != JsonToken.PROPERTY_NAME) {
            ix = p.nextNameMatch(matcher);
         } else {
            ix = p.currentNameMatch(matcher);
         }

         while (true) {
            while (ix < 0) {
               if (ix == -1) {
                  return bean;
               }

               if (ix != -2) {
                  return this._unexpectedToken(p, ctxt);
               }

               this._handleUnknown(p, ctxt, bean);
               ix = p.nextNameMatch(matcher);
            }

            SettableBeanProperty prop;
            Exception var10000;
            switch (ix) {
               case 0:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"imagesProp">();

                  try {
                     p.nextToken();
                     (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"imagesProp">().deserializeAndSet(p, ctxt, bean);
                  } catch (Exception var9) {
                     var10000 = var9;
                     boolean var10 = false;
                     break;
                  }

                  ix = p.nextNameMatch(matcher);
                  continue;
               case 1:
                  prop = (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"mediaProp">();

                  try {
                     if (p.nextToken() == JsonToken.START_OBJECT) {
                        (MethodHandle)GeneratedReaderBase.classDataEntry<"mediaSet">()
                           .invokeExact((Object)bean, (Object)(GeneratedReaderBase)GeneratedReaderBase.classDataEntry<"mediaCodec">().deserialize(p, ctxt));
                     } else {
                        (SettableBeanProperty)GeneratedReaderBase.classDataEntry<"mediaProp">().deserializeAndSet(p, ctxt, bean);
                     }
                  } catch (Exception var8) {
                     var10000 = var8;
                     boolean var10001 = false;
                     break;
                  }

                  ix = p.nextNameMatch(matcher);
                  continue;
               default:
                  throw new IllegalStateException("bad property index");
            }

            Exception e = var10000;
            throw this._propertyException(e, bean, prop, ctxt);
         }
      } else {
         return super._fallback.deserialize(p, ctxt);
      }
   }
}
```

Field-backed properties differ only in the store handle: a non-final public,
package-private, or private field is written through
`classDataEntry<"...Set">().invokeExact((Object)bean, (Object)value)` exactly
as a setter is (the handle is `Lookup.unreflectSetter` instead of
`unreflect`), so the arm shape is identical and no separate example is shown.
Final fields keep the stock property.
</details>

<details>
<summary><b>Serializer</b> - <code>Media</code> (scalar-rich writer, erased getter handles and padded helpers; shown in full, plus the bytecode view of one helper)</summary>

The generated writer: a straight-line sequence of per-kind helper calls with
pre-encoded name constants; each property value loads through an erased getter
handle (`(int)classDataEntry<"bitrateGet">().invokeExact((Object)value)`),
nested/list properties ride their stock writers. Every helper is padded past
the C2 inline threshold so its jackson-core hot calls compile as standalone
units (the AArch64 register-allocator finding in the PR text); decompilers
elide the padding, so the bytecode view below shows it.

```java
package tools.jackson.module.blackbird.codegen;

import java.lang.invoke.MethodHandle;
import tools.jackson.core.JsonGenerator;
import tools.jackson.core.SerializableString;
import tools.jackson.databind.SerializationContext;
import tools.jackson.databind.ser.PropertyWriter;
import tools.jackson.databind.ser.bean.BeanSerializerBase;
import tools.jackson.module.blackbird.internal.GeneratedWriterBase;

public final class BBWriter_Media extends GeneratedWriterBase {
   public BBWriter_Media(BeanSerializerBase fallback) {
      super(fallback);
   }

   public void serialize(Object value, JsonGenerator g, SerializationContext ctxt) {
      if (ctxt.getActiveView() == null) {
         g.writeStartObject(value);
         $int(g, (SerializableString)GeneratedWriterBase.classDataEntry<"bitrateName">(), (int)(MethodHandle)GeneratedWriterBase.classDataEntry<"bitrateGet">().invokeExact((Object)value));
         $str(g, (SerializableString)GeneratedWriterBase.classDataEntry<"copyrightName">(), (String)(Object)(MethodHandle)GeneratedWriterBase.classDataEntry<"copyrightGet">().invokeExact((Object)value));
         $long(g, (SerializableString)GeneratedWriterBase.classDataEntry<"durationName">(), (long)(MethodHandle)GeneratedWriterBase.classDataEntry<"durationGet">().invokeExact((Object)value));
         $str(g, (SerializableString)GeneratedWriterBase.classDataEntry<"formatName">(), (String)(Object)(MethodHandle)GeneratedWriterBase.classDataEntry<"formatGet">().invokeExact((Object)value));
         $int(g, (SerializableString)GeneratedWriterBase.classDataEntry<"heightName">(), (int)(MethodHandle)GeneratedWriterBase.classDataEntry<"heightGet">().invokeExact((Object)value));
         (PropertyWriter)GeneratedWriterBase.classDataEntry<"personsWriter">().serializeAsProperty(value, g, ctxt);
         (PropertyWriter)GeneratedWriterBase.classDataEntry<"playerWriter">().serializeAsProperty(value, g, ctxt);
         $long(g, (SerializableString)GeneratedWriterBase.classDataEntry<"sizeName">(), (long)(MethodHandle)GeneratedWriterBase.classDataEntry<"sizeGet">().invokeExact((Object)value));
         $str(g, (SerializableString)GeneratedWriterBase.classDataEntry<"titleName">(), (String)(Object)(MethodHandle)GeneratedWriterBase.classDataEntry<"titleGet">().invokeExact((Object)value));
         $str(g, (SerializableString)GeneratedWriterBase.classDataEntry<"uriName">(), (String)(Object)(MethodHandle)GeneratedWriterBase.classDataEntry<"uriGet">().invokeExact((Object)value));
         $int(g, (SerializableString)GeneratedWriterBase.classDataEntry<"widthName">(), (int)(MethodHandle)GeneratedWriterBase.classDataEntry<"widthGet">().invokeExact((Object)value));
         g.writeEndObject();
      } else {
         super._fallback.serialize(value, g, ctxt);
      }
   }

   private static void $str(JsonGenerator g, SerializableString name, String value) {
      g.writeName(name);
      if (value != null) {
         g.writeString(value);
      } else {
         g.writeNull();
      }
   }

   private static void $int(JsonGenerator g, SerializableString name, int value) {
      g.writeName(name);
      g.writeNumber(value);
   }

   private static void $long(JsonGenerator g, SerializableString name, long value) {
      g.writeName(name);
      g.writeNumber(value);
   }
}
```

The `$int` helper as bytecode - the 384 leading `nop`s are the inline-threshold
padding:

```
private static void $int(tools.jackson.core.JsonGenerator, tools.jackson.core.SerializableString, int);
    Code:
         0: nop
         1: nop
         2: nop
         3: nop
         // ... 380 more nop instructions (padding past FreqInlineSize=325) ...
       384: aload_0
       385: aload_1
       386: invokevirtual #135                // Method tools/jackson/core/JsonGenerator.writeName:(Ltools/jackson/core/SerializableString;)Ltools/jackson/core/JsonGenerator;
       389: pop
       390: aload_0
       391: iload_2
       392: invokevirtual #146                // Method tools/jackson/core/JsonGenerator.writeNumber:(I)Ltools/jackson/core/JsonGenerator;
       395: pop
       396: return
```
</details>

<details>
<summary><b>Polymorphic-subtype serializer</b> - <code>PolyEvent.MediaEvent</code> (native <code>serializeWithType</code> with the WritableTypeId flow)</summary>

A subtype of an `@JsonTypeInfo` hierarchy gets both `serialize` and a native
`serializeWithType` that replicates stock `BeanSerializerBase`'s flow exactly:
`typeSer.typeId(value, START_OBJECT)`, `writeTypePrefix`, `assignCurrentValue`,
the same property sequence as `serialize`, then `writeTypeSuffix`. Databind
calls `serializeWithType` on the subtype's serializer for every inclusion
mechanism (PROPERTY, WRAPPER_OBJECT, WRAPPER_ARRAY, ...), so the type wrapper
is written by the `TypeSerializer` and the body stays generated. An active
view, or a bean with a `@JsonTypeId` property, delegates to the stock path.
Only `serializeWithType` is shown; `serialize` is the same body without the
prefix/suffix pair.

```java
   public void serializeWithType(Object value, JsonGenerator g, SerializationContext ctxt, TypeSerializer typeSer) {
      if (ctxt.getActiveView() == null) {
         WritableTypeId typeIdDef = typeSer.typeId(value, JsonToken.START_OBJECT);
         typeSer.writeTypePrefix(g, ctxt, typeIdDef);
         g.assignCurrentValue(value);
         $long(g, (SerializableString)GeneratedWriterBase.classDataEntry<"durationName">(), (long)(MethodHandle)GeneratedWriterBase.classDataEntry<"durationGet">().invokeExact((Object)value));
         $str(g, (SerializableString)GeneratedWriterBase.classDataEntry<"formatName">(), (String)(Object)(MethodHandle)GeneratedWriterBase.classDataEntry<"formatGet">().invokeExact((Object)value));
         $str(g, (SerializableString)GeneratedWriterBase.classDataEntry<"nameName">(), (String)(Object)(MethodHandle)GeneratedWriterBase.classDataEntry<"nameGet">().invokeExact((Object)value));
         $bool(g, (SerializableString)GeneratedWriterBase.classDataEntry<"persistentName">(), (boolean)(MethodHandle)GeneratedWriterBase.classDataEntry<"persistentGet">().invokeExact((Object)value));
         $int(g, (SerializableString)GeneratedWriterBase.classDataEntry<"sizeName">(), (int)(MethodHandle)GeneratedWriterBase.classDataEntry<"sizeGet">().invokeExact((Object)value));
         typeSer.writeTypeSuffix(g, ctxt, typeIdDef);
      } else {
         super._fallback.serializeWithType(value, g, ctxt, typeSer);
      }
   }
```
</details>

### Reproducing

Run any Jackson 3 application with the module registered and `-Dblackbird.debug.dumpDir=/some/dir`: every generated reader and writer is written there as `<bean-class>-reader.class` / `<bean-class>-writer.class` before definition, one file per bean, ready for `javap` or any decompiler.
`-Dblackbird.debug.codegen=true` additionally traces generation and gate decisions to stderr. The examples above were decompiled with Vineflower 1.10.1.


cowtowncoder:
Ok: is JDK-25 hard requirement for building/releasing Blackbird module? JDK 21 already needed for building jackson-core 3.1+, everything else built with 17.

cowtowncoder:
## Code review (Claude Code)

I found 10 issues and one minor note, each checked against stock jackson-databind behavior. Nothing was run, and none of these cases is covered by the existing blackbird tests.

### High

1. **Beans that stock builds a special way go through the generated path anyway** (`BBReaderFactory.java:102`). The check is only `canCreateUsingDefault()`, which misses two cases:
   - An `EXTERNAL_PROPERTY` type id: stock `resolve()` removes that property from `_beanProperties`, so the codec sends both the property and its type id to the unknown-property handler. Reading fails with `UnrecognizedPropertyException`, or the value is silently dropped.
   - A delegating `@JsonCreator` alongside a no-arg constructor: the codec uses the constructor plus setters instead of the delegate.
2. **Polymorphic properties skip `deserializeWithType`** (`BBReaderFactory.java:378`, and the record path at `:311`). A property with a value type deserializer still becomes a CHILD arm and calls plain `deserialize()`. This breaks `@JsonTypeInfo(use=NAME)` on concrete bases and `activateDefaultTyping(NON_FINAL)`: the subtype is never built and the type id is read as an unknown property.

### Medium

3. **`@JsonManagedReference` is broken** (`BBReaderFactory.java:371`, `classify`). The store handle comes from `prop.getMember()` whatever the property's class is. `ManagedReferenceProperty` (and any `SettableBeanProperty.Delegating` wrapper) forwards `getMember()`, so its `deserializeAndSet` is bypassed and the back-reference (`child.parent`) stays null.
4. **Missing record components come out as null or 0** (`BeanReaderGenerator.java:793`). Stock `PropertyValueBuffer._findMissing` uses `NullValueProvider.getAbsentValue` and then `deser.getAbsentValue`. For example, `record R(String name, Optional<String> nick)` read from `{"name":"x"}` gives `Optional.empty()` in stock but `null` here. That null also wrongly trips `FAIL_ON_NULL_CREATOR_PROPERTIES`.
5. **Constructor and `build()` exceptions aren't wrapped** (`BeanReaderGenerator.java:414`, `:852`, `:624`). The no-arg constructor, the record canonical constructor and the builder `build()` are called through raw `invokeExact`:
   - A record whose compact constructor throws `IllegalArgumentException` leaks that raw exception instead of `ValueInstantiationException`.
   - `DeserializationProblemHandler.handleInstantiationProblem` is never called for beans that have properties.
   - The existing `BustedCtor` test only passes because that bean has no properties and stays on the stock path.
6. **Names ignored because of `@JsonIgnoreType` are lost** (`BBDeserializerModifier.java:149`). The codec's ignore list is rebuilt from annotations and misses names the stock factory adds with `builder.addIgnorable` (`BeanDeserializerFactory.filterBeanProps`). JSON containing those properties now fails in `handleUnknownProperty` instead of being skipped. Fix: read `_ignorableProps`, `_includableProps` and `_ignoreAllUnknown` from the resolved stock deserializer.
7. **Custom null serializers are ignored** (`BBWriterFactory.java:118`). STRING and CHILD arms write `writeNull()` directly, bypassing the property's `_nullSerializer` (for example `@JsonSerialize(nullsUsing=...)`).
8. **The CHILD writer arm skips three things stock `BeanPropertyWriter.serializeAsProperty` does** (`BBWriterFactory.java:143`):
   - The property's `_typeSerializer`, so type ids are dropped under `USE_STATIC_TYPING` with `@JsonTypeInfo`, or with final polymorphic types.
   - The self-reference check, so `FAIL_ON_SELF_REFERENCES` becomes a `StackOverflowError`.
   - `_applyView`.
9. **Generated `serialize` / `serializeWithType` lose error wrapping** (`BeanWriterGenerator.java:307`, `:369`). There is no equivalent of stock's `catch (Exception)` → `wrapAndThrow` or `catch (StackOverflowError)` → "Infinite recursion". A throwing getter escapes as a raw (possibly checked) exception with no path, and a cyclic object graph ends in a raw `StackOverflowError` instead of a `DatabindException`.
10. **Blackbird now requires Java 25** (`blackbird/pom.xml:30`). It compiles at release 25 for the ClassFile API and only builds under the JDK 25+ profile, while the project baseline is Java 17:
    - Users on JDK 17–24 who upgrade get `UnsupportedClassVersionError`.
    - With the jar on the classpath of a JDK 17–24 app that uses `findModules()`, `ServiceLoader` iteration fails with `ServiceConfigurationError`, which breaks auto-registration of *every* module.

### Low

- `module-info.java:12` now exports `tools.jackson.module.blackbird.internal`. That goes against the project's rule not to export internal packages, and it looks unnecessary because the hidden classes are defined inside blackbird's own module.

### Summary

The most urgent are #10, which affects every user below Java 25, and #1–2, which lose data or fail on reads that work today. Most of the rest come from the generated code skipping things stock `BeanPropertyWriter` / `SettableBeanProperty` do. Falling back to the stock path whenever a property has a type deserializer or serializer, a custom null handler, or a wrapper property class would close most of them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

